PR #28, opened 2015-04-21
Fix/postgresql deadlocks

HonoreDB:
Merges feature/postgres into master, and strengthens the lock mode to reduce deadlocks. `share mode` appears to be a mistake according to postgres docs:

>If a transaction of this sort is going to change the data in the table, then it should use SHARE ROW EXCLUSIVE lock mode instead of SHARE mode. This ensures that only one transaction of this type runs at a time. Without this, a deadlock is possible: two transactions might both acquire SHARE mode, and then be unable to also acquire ROW EXCLUSIVE mode to actually perform their updates.

Tests pass.

### DIFF
--- a/lib/policy_machine.rb
+++ b/lib/policy_machine.rb
@@ -163,9 +163,9 @@ class PolicyMachine
   # object (attribute).
   #
   # TODO:  might make privilege a class of its own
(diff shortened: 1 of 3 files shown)


yzhangmedidata:
Very interesting find, I'll deploy a branch in dalton with this PM branch and check the changes

PR #31, opened 2015-05-04
Fix active_record has_and_belongs_to_many performance

HonoreDB:
Betrayed! Betrayed by ActiveRecord!

If you have an operation set with three elements, and you change it to an operation set with 111 elements, the innocuous command `operations=` actually generates 108 inserts (and for some reason also 108 queries to table metadata). This does not appear to be something ActiveRecord can be talked out of doing. So I've overwritten the method to take the diff and convert it into (at most) two writes, a bulk delete and bulk insert.

I've been trying to keep raw SQL out of the main active_record adapter, so I brought in a gem that adds the `.import` method to ActiveRecord classes to do the bulk insert bit. Tests pass with both postgres and mysql.

Future plans include cleaning this (and the extra_attributes nonsense) up in the Postgres adapter using collection columns (hstore, array, iarray).

### DIFF
--- a/lib/policy_machine_storage_adapters/active_record.rb
+++ b/lib/policy_machine_storage_adapters/active_record.rb
@@ -3,7 +3,7 @@ require 'set'
 
 # This class stores policy elements in a SQL database using whatever
 # database configuration and adapters are provided by active_record.
-# Currently only MySQL is supported via this adapter.
+# Currently only MySQL and Postgres are supported via this adapter.
 
 begin
   require 'active_record'
@@ -14,6 +14,8 @@ end
 module PolicyMachineStorageAdapter
 
   class ActiveRecord
+
+    require 'activerecord-import' # Gem for bulk inserts
 
     # Load the Assignment class at runtime because it's implemented differently by different adapters
     # And which database adapter is active is not always determinable at class definition time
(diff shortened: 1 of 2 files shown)


yzhangmedidata:
Should we do sanity checks of dalton with this branch before merging into master

policy_machine.gemspec
```diff
@@ -23,5 +23,6 @@ Gem::Specification.new do |s|
   s.add_development_dependency('database_cleaner')
   s.add_development_dependency('will_paginate', '~> 3.0.5')
   s.add_development_dependency('activerecord-hierarchical_query', '~> 0.0')
+  s.add_development_dependency('activerecord-import', '~> 0.0')
```
csavage-mdsol: Should we really not add this as a runtime dependency? I mean, if you decide to use another backend, you wont need it, but figuring out what dependencies a gem actually needs in an extra bundler fashion sounds slightly annoying.  Maybe a doc update? Oh, ditto for hierarchical_query above this as well.

lib/policy_machine_storage_adapters/active_record.rb
```diff
@@ -111,6 +113,27 @@ class PolicyElementAssociation < ::ActiveRecord::Base
...
+        current_operation_set = Set.new(self.operations)
+        new_operations = updated_operation_set - current_operation_set
+        removed_operations = current_operation_set - updated_operation_set
+        transaction do
+          OperationsPolicyElementAssociation.where(policy_element_association_id: self.id)
+                                            .where(operation_id: removed_operations.map(&:id))
+                                            .delete_all
+          OperationsPolicyElementAssociation.import([:policy_element_association_id, :operation_id],
+                                                     new_operations.map{ |op| [self.id, op.id] },
+                                                     validate: false)
```
csavage-mdsol: Whats up with the ```validate: false```?

policy_machine.gemspec
HonoreDB: Yeah, I agree it's getting annoying. Actually, reading the docs it looks like runtime dependencies in a gemspec don't get autorequired, so there's little reason not to add them as runtime. Will change.

lib/policy_machine_storage_adapters/active_record.rb
HonoreDB: There are no validations, I was just worried the gem would instantiate model elements if I didn't pass that in.

HonoreDB:
@csavage-mdsol Want me to re-open this pointed at develop?

csavage-mdsol:
@HonoreDB ooh, yes please.  If you assert specs pass in that new PR, will merge

PR #32, opened 2015-05-05
Allow ignore_case to specify which attributes to ignore case on

HonoreDB:
One more performance optimization: we're currently generating SQL that does case insensitive comparisons inappropriately, including on the `policy_machine_uuid` column (which should never happen) and on user-created uuid attribute columns (which should be under the user's control), preventing the use of indexes.

This PR modifies the in-memory and active_record backends so you can do `find_all_of_type_object(name: 'Eric', color: 'red', ignore_case: [:name])`. Case insensitivity was never implemented for neo4j.


mszenher:
@HonoreDB Do all specs pass?

HonoreDB:
They do.

PR #35, opened 2015-05-06
Fix/active record habtm performance

HonoreDB:
Specs pass.

Finished in 14.21 seconds
571 examples, 0 failures

(See discussion at #31)

### DIFF
--- a/lib/policy_machine_storage_adapters/active_record.rb
+++ b/lib/policy_machine_storage_adapters/active_record.rb
@@ -3,7 +3,7 @@ require 'set'
 
 # This class stores policy elements in a SQL database using whatever
 # database configuration and adapters are provided by active_record.
-# Currently only MySQL is supported via this adapter.
+# Currently only MySQL and Postgres are supported via this adapter.
 
 begin
   require 'active_record'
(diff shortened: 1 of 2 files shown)


lib/policy_machine_storage_adapters/active_record.rb
```diff
@@ -111,6 +113,27 @@ class PolicyElementAssociation < ::ActiveRecord::Base
       has_and_belongs_to_many :operations, class_name: :"PolicyMachineStorageAdapter::ActiveRecord::Operation"
       belongs_to :user_attribute
       belongs_to :object_attribute
+
+      #TODO: ActiveRecord's generated operations= method is inefficient, makes 1 query for each op added or removed even though there's no hooks
+      # Awkward manual implementation for now, but in the future change this to an hstore or something in the postgres adapter,
+      # and/or fix Rails.
+      def operations=(updated_operations)
```
yzhangmedidata: Should there actually be additional specs to make sure this method behaves correctly

HonoreDB: That hasn't generally been our style with specs here, but if you're feeling nervous about this method, feel free to insist.

yzhangmedidata: Merging as is then

PR #38, opened 2015-05-29
[MCC-168375] removes cycle checking for performance

drozenberg:
- Remove .no_cycle method when recursively querying for decendants
- Yields ~10% performance increase based on benchmark tests

@yzhangmedidata @HonoreDB @csavage-mdsol

### DIFF
--- a/lib/policy_machine_storage_adapters/active_record/postgresql.rb
+++ b/lib/policy_machine_storage_adapters/active_record/postgresql.rb
@@ -15,14 +15,14 @@ module PolicyMachineStorageAdapter
 
       def self.descendants_of(element)
         recursive_query = join_recursive do |query|
-          query.start_with(parent_id: element.id).connect_by(child_id: :parent_id).nocycle
+          query.start_with(parent_id: element.id).connect_by(child_id: :parent_id)
         end
         PolicyElement.where(id: recursive_query.select('assignments.child_id'))
       end
 
       def self.ancestors_of(element)
         recursive_query = join_recursive do |query|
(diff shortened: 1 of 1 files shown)


yzhangmedidata:
The standard procedure for these PM changes has been to update Dalton to use this branch of the PM, deploy it to sandbox and ensure that performance metrics in NewRelic matches expectations, once this is verified this can be merged

HonoreDB:
10%, good to know! I wasn't sure whether to root for a substantial performance improvement here, since it sort of obligates us to maintain the A of the DAG, but I think that's a reasonable sacrifice to make.

drozenberg:
@yzhangmedidata @HonoreDB is this an acceptable way to convert the rake task to a migration? we can deploy this and see if dalton stays up this time.

agodel:
@HonoreDB @yzhangmedidata @csavage-mdsol  we were successful in deploying a branch to dalton that points to this branch, cycles are fixed, does that make this ready to merge?

yzhangmedidata:
What was the final performance metrics differences with this branch on sandbox?

agodel:
DO NOT MERGE YET

@yzhangmedidata it's hard to say exactly, it looks like postrgres select is slightly faster on privilege show via new relic but since traffic is different it's hard to say. We need to not merge this into stable-0.1 and instead make it it's own branch. 'stable-0.1.1' potentially? or 'stable-0.2'. Otherwise we could mess up validation.

yzhangmedidata:
This should be closed and retargeted to the stable-0.2 branch then

PR #39, opened 2015-06-01
Add global config for whether to tolerate cycles

HonoreDB:
@yzhangmedidata @agodel @drozenberg I hadn't bothered with this because I didn't see a use case, but it sounds like this could make a two-phased deployment significantly easier, so here it is if you want it.

Makes cycle tolerance globally configurable since our dependency structure makes configuring a specific adapter awkward. As it happens, it looks like all adapters other than Postgres tolerate cycles just fine (I'm not sure how neo4j is managing it, but it is).

Specs pass.

### DIFF
--- a/spec/support/shared_examples_policy_machine.rb
+++ b/spec/support/shared_examples_policy_machine.rb
@@ -370,6 +370,17 @@ shared_examples "a policy machine" do
       policy_machine.is_privilege?(@u1, @w, @o1).should be_false
     end
 
+    context 'tolerate_cycles is set to true' do
+      before { PolicyMachine.config[:tolerate_cycles] = true }
+      after  { PolicyMachine.config.clear }
+      it 'tolerates cycles when configured' do
+        @groupa = policy_machine.create_user_attribute('GroupA')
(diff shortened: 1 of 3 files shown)


spec/support/shared_examples_policy_machine.rb
```diff
@@ -370,6 +370,17 @@
       policy_machine.is_privilege?(@u1, @w, @o1).should be_false
     end
 
+    context 'tolerate_cycles is set to true' do
```
yzhangmedidata: What about a test for the opposite, <code>tolerate_cycles</code> set to false

HonoreDB: The problem is that the expected behavior is works-as-normal for most adapters, hang-and-create-a-zombie in Postgres.

yzhangmedidata: Can the postgres adapter be excluded with a conditional but have a test for the mysql and in memory adapters

HonoreDB:
Closing because the performance benefit doesn't appear to be high enough to be worth pushing this.

PR #27, opened 2015-03-25
Support prohibitions

HonoreDB:
Couldn't help myself. Don't worry about reviewing this right now, we may not be consuming it yet.

This implements prohibitions at the policy machine level, without needing to touch the storage adapters, by storing them as operations. This detail is hidden from the consumer so we can change it later if we want.

Parallelizing the prohibition check should be doable, but it'll require some changes to the adapters so I've left it as a TODO.

PR #40, opened 2015-06-10
Feature/prohibitions as negated operations

HonoreDB:
See #27 and its thread.

Specs continue to pass.

### DIFF
--- a/spec/support/shared_examples_policy_machine.rb
+++ b/spec/support/shared_examples_policy_machine.rb
@@ -245,6 +245,10 @@ shared_examples "a policy machine" do
       expect(PM::Prohibition.on(:fly)).to be_a Symbol
     end
 
+    it 'can negate operations expressed as PM::Operations' do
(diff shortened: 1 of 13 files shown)


spec/support/shared_examples_policy_machine.rb
```diff
@@ -220,6 +220,33 @@
...
+
+    it 'raises if trying to negate a non-operation' do
+      expect{PM::Prohibition.on(3)}.to raise_error(ArgumentError)
+    end
+
+    it 'can negate operations expressed as strings' do
+      expect(PM::Prohibition.on('fly')).to be_a String
+    end
+
+    it 'can negate operations expressed as symbols' do
```
yzhangmedidata: Should there be another test that it can negate operations of class Operation per https://github.com/mdsol/the_policy_machine/pull/40/files#diff-cd736b91a0f9140de3bb79b73ec78b70R308

yzhangmedidata:
Looks good

PR #42, opened 2015-09-03
fixes maudit-client bug by defining methods only on PE and not base

agodel:
The purpose of this PR is to fix a bug cause by this line. Without this change, if there is any attribute stored in "extra_attributes" that shares the name of an attribute on another model in the application that inherits directly from ActiveRecord::Base, that model will expect that attribute to be stored in "extra_attributes" as well because of this. Targeting the method at PolicyElement fixes this.

@mdsol/team04 @csavage-mdsol @DMcKinnon-mdsol


yzhangmedidata:
Looks good

DMcKinnon-mdsol:
Just to be clear – is this ever something that another model might encounter using something similar to "extra_attributes"? Does it make sense to have a spec around this?

HonoreDB:
@DMcKinnon-mdsol Not with any sane way of writing it. If someone else had this bug or otherwise messed with the `method_missing` attribute on AR::Base, you might hit it, but I don't know if you really want an anti-monkeypatching spec...that pulls you into speccing out all sorts of behavior you want to take for granted.

DMcKinnon-mdsol:
Cool. Thanks!

PR #43, opened 2015-09-03
adds ability to pass a custom where clause to the storage adapter if it is an active record db

agodel:
The purpose of this PR is to support custom where clauses for the active_record storage adapter. If we pass :where to the find all method, it will insert a where clause with the value passed"

@mdsol/team04 

This is necessary functionality for nested config type roles to be performant in Dalton

### DIFF
--- a/lib/policy_machine_storage_adapters/in_memory.rb
+++ b/lib/policy_machine_storage_adapters/in_memory.rb
@@ -28,6 +28,10 @@ module PolicyMachineStorageAdapter
       # The results are paginated via will_paginate using the pagination params in the params hash
       # The find is case insensitive to the conditions
       define_method("find_all_of_type_#{pe_type}") do |options = {}|
+        if options.delete(:where)
+          raise NotImplementedError.new('you are trying to use an active_record where clause on an in memory storage adapter!')
+        end
+
         conditions = options.slice!(:per_page, :page, :ignore_case).merge(pe_type: pe_type)
         elements = policy_elements.select do |pe|
           conditions.all? do |k,v|
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+# 0.0.3
+
+* Adds functionality for custom where clauses using activerecord storage adapter
+
 # 0.0.2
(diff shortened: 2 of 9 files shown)


agodel:
@HonoreDB does this change necessitate a version change?

HonoreDB:
It's a new feature, so should be a minor version bump.

lib/policy_machine_storage_adapters/in_memory.rb
```diff
@@ -28,6 +28,10 @@ class InMemory
       # The results are paginated via will_paginate using the pagination params in the params hash
       # The find is case insensitive to the conditions
       define_method("find_all_of_type_#{pe_type}") do |options = {}|
+        if options.delete(:where)
```
HonoreDB: You need this in the neography adapter too. Also, this is going to be awkward for integration testing in in-memory mode. If you'd like, you could make this more limited (an `attribute_containing` key or something like that) and I could add it to the other adapters. Part of the reason we don't expose active_record functionality directly is to maintain compatibility with non-sql graph databases.

agodel: make sense but can you clarify "If you'd like, you could make this more limited (an attribute_containing key or something like that) and I could add it to the other adapters" ? I understand not wanting to expose activerecord functionality for that reason, but shouldn't it be up to the consumers (i.e. dalton) to manage this if it is useful to have such functionality?

HonoreDB: So long as we limit the functionality, it'll require a finite amount of work to maintain feature parity across adapters. This is immediately useful in Dalton, since it uses two different adapters, and will also be helpful in the future when we (likely) move away from SQL in Dalton. If we allow custom `where`s now, then in a year that migration will be harder.

agodel:
@mdsol/team04 added the invalid tag to demonstrate that this isn't for merging yet, will get nested CTR functionality working then finish PR.

CGDS:
...and this one?

HonoreDB:
The purge concludes.

PR #44, opened 2015-09-08
[DO NOT MERGE] adds functionality to serialize CTR role uuids array

agodel:
This is a test branch I will be pointing dalton at to support necessary functionality of using array columns in rails. We want to abstract this logic such that it does not need to know about dalton specific columns but rather handles array columns correctly.


drozenberg:
LGTM

CGDS:
...I'm guessing we can close this one?

HonoreDB:
Yup, nothing's pointing at this anymore.

PR #46, opened 2015-09-17
swap out debugger for pry to make multiple ruby version use easy

csavage-mdsol:
@HonoreDB it's time

```
Finished in 26.35 seconds
593 examples, 0 failures

Randomized with seed 8512
```


HonoreDB:
Man, now I have to learn a thing.

PR #48, opened 2015-11-06
Feature/rails4 includes

HonoreDB:
Adds support for find_all_of_type_foo(tags: {include: 'up'}). Works if tags is an array or a string. Tests will pass with upcoming PR to the test gem.


CGDS:
@HonoreDB I've merged this into my branch locally.  One of the shared new specs is failing, but Im taking a look at it now.  Also, I did the generate a new project thing, no more external test gem.  Closing for now, will push soon.

PR #49, opened 2015-11-09
Tolerate string keys

HonoreDB:
Hash keys may be strings (as in a HashWithIndifferentAccess, where .keys always returns a string form even if they started out as symbols). Tolerate that.


agodel:
nice

agodel:
@HonoreDB @CGDS should I merge now? do we need this in the stable branch

PR #51, opened 2016-01-19
Add accessible objects list per user and operation

HonoreDB:
Adds PolicyMachine#accessible_objects method, using the same "slow default implementation, fast(ish) implementation in the activerecord adapter"). 

The mysql adapter seems to have been broken by the Rails update, so I wasn't able to add it to this. It should be a simple change, though. I'll fix mysql in a separate PR.

I haven't performance tested this yet, but if I end up optimizing it further I don't think it'll change the API, which is just `policy_machine.accessible_objects(user_or_user_attribute, operation, [includes: substring], [ignore_prohibitions: true])` returning an array of internal representations of objects.

Specs:

Finished in 2 minutes 18.76 seconds
931 examples, 0 failures, 22 pending
Coverage report generated for RSpec to /Users/aweiner/the_policy_machine/coverage. 776 / 780 LOC (99.49%) covered.

### DIFF
--- a/lib/policy_machine.rb
+++ b/lib/policy_machine.rb
@@ -69,7 +69,7 @@ class PolicyMachine
   # TODO: Parallelize the two component checks
   def is_privilege?(user_or_attribute, operation, object_or_attribute, options = {})
     is_privilege_ignoring_prohibitions?(user_or_attribute, operation, object_or_attribute, options) &&
-      !is_privilege_ignoring_prohibitions?(user_or_attribute, PM::Prohibition.on(operation), object_or_attribute, options)
+      (options[:ignore_prohibitions] || !is_privilege_ignoring_prohibitions?(user_or_attribute, PM::Prohibition.on(operation), object_or_attribute, options))
   end
 
   ##
@@ -191,6 +191,21 @@ class PolicyMachine
     else
       prohibited_operations = prohibitions.map { |_,prohibition,_| prohibition.operation }
       privileges.reject { |_,op,_| prohibited_operations.include?(op.unique_identifier) }
+    end
+  end
+
+  ##
+  # Returns an array of all objects the given user (attribute)
+  # has the given operation on.
+  def accessible_objects(user_or_attribute, operation, options = {})
+    if policy_machine_storage_adapter.respond_to?(:accessible_objects)
+      policy_machine_storage_adapter.accessible_objects(user_or_attribute, operation, options)
+    else
+      result = objects.select { |object| is_privilege?(user_or_attribute, operation, object, options) }
+      if inclusion = options[:includes]
+        result.select! { |object| object.unique_identifier.include?(inclusion) }
+      end
+      result
     end
   end
 
--- a/lib/policy_machine_storage_adapters/active_record.rb
+++ b/lib/policy_machine_storage_adapters/active_record.rb
@@ -345,16 +345,41 @@ module PolicyMachineStorageAdapter
       end
     end
 
-    private
-
-    def is_privilege_single_policy_class(user_or_attribute, operation, object_or_attribute)
-      if operation.is_a?(class_for_type('operation'))
-        associations_between(user_or_attribute, object_or_attribute).where(id: operation.policy_element_associations).any?
+    ## Optimized version of PolicyMachine#scoped_privileges
+    # Returns all objects the user has the given operation on
+    # TODO: Support multiple policy classes here
+    def accessible_objects(user_or_attribute, operation, options = {})
+      operation = class_for_type('operation').find_by_unique_identifier!(operation.to_s) unless operation.is_a?(class_for_type('operation'))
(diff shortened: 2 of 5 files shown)


HonoreDB:
ping @agodel or @CGDS

lib/policy_machine.rb
```diff
@@ -69,7 +69,7 @@ def add_association(user_attribute_pe, operation_set, object_attribute_pe)
   # TODO: Parallelize the two component checks
   def is_privilege?(user_or_attribute, operation, object_or_attribute, options = {})
     is_privilege_ignoring_prohibitions?(user_or_attribute, operation, object_or_attribute, options) &&
-      !is_privilege_ignoring_prohibitions?(user_or_attribute, PM::Prohibition.on(operation), object_or_attribute, options)
+      (options[:ignore_prohibitions] || !is_privilege_ignoring_prohibitions?(user_or_attribute, PM::Prohibition.on(operation), object_or_attribute, options))
```
CGDS: Why did we need this one?

HonoreDB: Oh, that might actually be a legacy of an earlier implementation. Let me see if I can remove it.

HonoreDB: Oh, right, no, that's to support the default implementation that uses is_privilege? (line 204). Easiest way to have it support ignore_prohibitions was to pass it through to is_privilege?.

lib/policy_machine_storage_adapters/active_record.rb
```diff
@@ -345,16 +345,41 @@ def scoped_privileges(user_or_attribute, object_or_attribute, options = {})
       end
     end
 
+    ## Optimized version of PolicyMachine#scoped_privileges
+    # Returns all objects the user has the given operation on
+    # TODO: Support multiple policy classes here
+    def accessible_objects(user_or_attribute, operation, options = {})
+      operation = class_for_type('operation').find_by_unique_identifier!(operation.to_s) unless operation.is_a?(class_for_type('operation'))
+      permitting_oas = PolicyElement.where(id: operation.policy_element_associations.where(
```
CGDS: whats the `to_sql` on this one look like?  iirc AR handles inner `select`s right, but not sure if thats true when chained with a `where`.

HonoreDB:
```sql
SELECT "policy_elements".* FROM "policy_elements" WHERE "policy_elements"."id" IN (SELECT "policy_element_associations"."object_attribute_id" FROM "policy_element_associations" INNER JOIN "operations_policy_element_associations" ON "policy_element_associations"."id" = "operations_policy_element_associations"."policy_element_association_id" WHERE "operations_policy_element_associations"."operation_id" = 4 AND "policy_element_associations"."user_attribute_id" IN (7, 6))
```

So two calls, I think, one for `.descendants` and one for the rest. I think that's because `|` on a chain isn't working right anymore, that's a problem with at least one other method in this file too.

lib/policy_machine.rb
CGDS: ah, gotcha

lib/policy_machine_storage_adapters/template.rb
```diff
@@ -173,5 +173,11 @@ def scoped_privileges(user_or_attribute, object_or_attribute)
 
     end
 
+    # Optimized version of PolicyMachine#accessible_objects
+    # Return all objects the user has the given operation on
+    # Optional: only add this method if you can do it better than policy_machine.rb
+    def accessible_objects(user_or_attribute, operation, options = {})
```
CGDS: Not related to outstanding pr, but why did we do this instead of inheritance + `NotImplemented` errors again?

HonoreDB: I think it might've been Matt's idea--I want to blame it on him but the git history of this file doesn't work in my favor. I guess the point was that copying a file and filling in method definitions was easier with a template than with a module, and superclasses would prevent inheriting from like a Datamapper class or something. I've never done it this way before or since though.

lib/policy_machine_storage_adapters/active_record.rb
CGDS: It feels like there should be some sneaky `or` thing or casting a record back to a relation thingamajig AR would provide that would let you sneak out of this, but I can't think of one at the moment.  Oh well.

CGDS: Oh, and do you think `|` is forcing the `to_a` call?  Maybe I was thinking of squeel when I was trying to be sneaky.

CGDS:
Looks good, will merge by end of day unless further comments

lib/policy_machine_storage_adapters/active_record.rb
HonoreDB: Yeah, I think squeel would do it.

PR #52, opened 2016-02-11
allow passing in column name to seach by as key

audibleblink:
This PR allows PM users to pass in the column name that is to be searched for the purpose of deriving bulk permissions.

### DIFF
--- a/lib/policy_machine.rb
+++ b/lib/policy_machine.rb
@@ -144,7 +144,7 @@ class PolicyMachine
     end
   end
 
-  
+
 
   ##
   # Returns an array of all privileges encoded in this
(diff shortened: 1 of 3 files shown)


CGDS:
@audibleblink do you think we were missing some test coverage for this one, or do you think we need to add some more tests for our newfound freedom?

audibleblink:
@CGDS good question. should we be telling users that options[:key] is required? If one forgets to pass it in, they're left with an unhelpful error message.

In any case, modified our current specs to account for the new signature requirements.

audibleblink:
```
Finished in 17.71 seconds
677 examples, 0 failures, 12 pending
Coverage report generated for RSpec to /Users/eflores/Code/Medidata/the_policy_machine/coverage. 699 / 780 LOC (89.62%) covered.
```

PR #59, opened 2016-06-09
[MCC-233516] - add find_in_batches-like functiniality

audibleblink:
This PR adds the ability to search policy_elements in batches with the postgres adapter

### DIFF
--- a/lib/policy_machine.rb
+++ b/lib/policy_machine.rb
@@ -195,6 +195,17 @@ class PolicyMachine
   end
 
   ##
+  # Search for and iterate over a collection in batches
+  def batch_find(type:, query: {}, config: {}, &blk)
+    if policy_machine_storage_adapter.respond_to?(:batch_find)
+      policy_machine_storage_adapter.batch_find(type, query, config, &blk)
+    else
+      batch_size = config.fetch(:batch_size, 1)
+      method(type.to_s.pluralize).call(query).each_slice(batch_size, &blk)
+    end
+  end
+
+  ##
   # Returns an array of all objects the given user (attribute)
   # has the given operation on.
   def accessible_objects(user_or_attribute, operation, options = {})
--- a/spec/support/shared_examples_policy_machine.rb
+++ b/spec/support/shared_examples_policy_machine.rb
@@ -842,6 +842,12 @@ shared_examples "a policy machine" do
 
     context 'when given a block' do
 
+      it 'calls the block' do
+        expect do |spy|
+          policy_machine.batch_find(type: :object, query: { unique_identifier: 'one:fish' }, &spy)
+        end.to yield_control
+      end
+
       context 'and search terms' do
         it 'returns the matching records' do
           policy_machine.batch_find(type: :object, query: { unique_identifier: 'one:fish' }) do |batch|
@@ -855,6 +861,10 @@ shared_examples "a policy machine" do
         it 'returns the correct batch size' do
           policy_machine.batch_find(type: :object, config: { batch_size: 1 }) do |batch|
             expect(batch.size).to eq 1
+          end
+
+          policy_machine.batch_find(type: :object, config: { batch_size: 3 }) do |batch|
+            expect(batch.size).to eq 3
           end
         end
       end
@@ -889,9 +899,9 @@ shared_examples "a policy machine" do
 
       context 'but given config options' do
         it 'resepects batch size configs while return all results' do
-          enum = policy_machine.batch_find(type: :object, config: { batch_size: 1})
+          enum = policy_machine.batch_find(type: :object, config: { batch_size: 3})
           results = enum.flat_map do |batch|
-            expect(batch.size).to eq 1
+            expect(batch.size).to eq 3
             batch.map { |pe| pe.unique_identifier }
           end
           expected = %w(one:fish two:fish red:one)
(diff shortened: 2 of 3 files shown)


audibleblink:
opening now for discussion/transparency. thoughts @mdsol/team-04 ?

DMcKinnon-mdsol:
Have you tested this out and looked at the actual sql calls it's making? Is it limiting each request to 1000 records correctly?

audibleblink:
I have, batch sizing works too. 

```
D, [2016-06-09T13:08:04.896810 #11236] DEBUG -- :   PolicyMachineStorageAdapter::ActiveRecord::UserAttribute Load (0.9ms)  SELECT  "policy_elements".* FROM "policy_elements" WHERE "policy_elements"."type" IN ('PolicyMachineStorageAdapter::ActiveRecord::UserAttribute') AND ("policy_elements"."id" > 11)  ORDER BY "policy_elements"."id" ASC LIMIT 5
```

HonoreDB:
Ooh, that's so elegant, I love that the scope chaining just works.

lib/policy_machine.rb
```diff
@@ -195,6 +195,16 @@ def scoped_privileges(user_or_attribute, object_or_attribute, options = {})
   end
 
   ##
+  # Search for and iterate over a collection in batches
+  def batch_find(policy_object, query = {}, config = {}, &blk)
+    if policy_machine_storage_adapter.respond_to?(:batch_find)
+      policy_machine_storage_adapter.batch_find(policy_object, query, config, &blk)
+    else
+      raise 'not implemented'
```
HonoreDB: raise NotImplementedError for style points

HonoreDB: (or implement it for more style points :p )

HonoreDB:
API+implementation looks great to me, just needs specs when you're ready as far as I'm concerned.

audibleblink:
```rb
Finished in 16.83 seconds
695 examples, 0 failures, 12 pending
Coverage report generated for RSpec to /Users/eflores/Code/Medidata/the_policy_machine/coverage. 21 / 21 LOC (100.0%) covered.
```

lib/policy_machine.rb
```diff
@@ -195,6 +195,20 @@ def scoped_privileges(user_or_attribute, object_or_attribute, options = {})
   end
 
   ##
+  # Search for and iterate over a collection in batches
+  def batch_find(type:, query: {}, config: {}, &blk)
+    if policy_machine_storage_adapter.respond_to?(:batch_find)
+      policy_machine_storage_adapter.batch_find(type, query, config, &blk)
+    else
+      method(type.to_s.pluralize).call.select do |pm_el|
```
HonoreDB: Would `method(type.to_s.pluralize)`.call(query)` work instead of the select? Also, this looks like it doesn't handle the block, and assumes a batch size of 1. I think something like `.each_slice(config[:batch_size] || 1, &blk)` would work, it'll still return an enumerator if there is no block.

audibleblink: Oh rad. `call(query)` work because of [this](https://github.com/mdsol/the_policy_machine/blob/b9e6dec8bf6da3e20bc9d08471f72ee1472dd6b0/lib/policy_machine.rb#L261) line?

audibleblink:
```sh
Finished in 18.82 seconds
695 examples, 0 failures, 12 pending
```

spec/support/shared_examples_policy_machine.rb
```diff
@@ -820,4 +820,85 @@
...
+        it 'returns the matching records' do
+          policy_machine.batch_find(type: :object, query: { unique_identifier: 'one:fish' }) do |batch|
+            expect(batch.size).to eq 1
+            expect(batch.first.unique_identifier).to eq 'one:fish'
+          end
+        end
+      end
+
+      context 'and config options' do
+        it 'returns the correct batch size' do
```
HonoreDB: Now that you're honoring batch size in the default implementation, mind passing in a batch size > 1 to document that?

spec/support/shared_examples_policy_machine.rb
```diff
@@ -820,4 +820,85 @@
...
+          batch.map { |pe| pe.unique_identifier }
+        end
+          expected = 'one:fish'
+          expect(results.first).to eq(expected)
+        end
+      end
+
+      context 'but given config options' do
+        it 'resepects batch size configs while return all results' do
+          enum = policy_machine.batch_find(type: :object, config: { batch_size: 1})
```
HonoreDB: Here too, please, batch size greater than 1.

spec/support/shared_examples_policy_machine.rb
```diff
@@ -820,4 +820,85 @@
...
+      [@one_fish, @two_fish, @red_one].each do |object|
+        policy_machine.add_association(@ua, Set.new([@read]), object)
+      end
+      @oa = policy_machine.create_object_attribute('oa')
+      policy_machine.add_association(@ua, Set.new([@write]), @oa)
+      policy_machine.add_assignment(@u1, @ua)
+      policy_machine.add_assignment(@red_one, @oa)
+    end
+
+    context 'when given a block' do
```
HonoreDB: This should also have a test that [asserts that the block is called](https://www.relishapp.com/rspec/rspec-expectations/docs/built-in-matchers/yield-matchers). Otherwise since all your assertions are inside the passed-in block, a method that didn't do anything would still pass all the specs.

audibleblink:
```
Finished in 19.21 seconds
698 examples, 0 failures, 12 pending
```

HonoreDB:
Awesome, will merge after lunch if no more comments.

PR #60, opened 2016-06-14
rebases feature/serializers branch

audibleblink:
this pr rebases serializers back into develop along with the new batch_find method. grant cucumber tests fail without it.

### DIFF
--- a/lib/policy_machine.rb
+++ b/lib/policy_machine.rb
@@ -195,6 +195,17 @@ class PolicyMachine
   end
 
   ##
+  # Search for and iterate over a collection in batches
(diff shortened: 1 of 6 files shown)


HonoreDB:
It seems like @DMcKinnon-mdsol has found that this and `includes: ` don't play well together, probably because of https://github.com/mdsol/the_policy_machine/blob/156c8920d06349909f1f4c0fa399e3c7df631b45/lib/policy_machine_storage_adapters/active_record.rb#L168. Is there a simple way we can update that?

HonoreDB:
From conversations elsewhere, no. Merging as-is to unblock other work, let's not tag yet though.

PR #62, opened 2016-07-25
Feature/generic batch writes

HonoreDB:
Adds the ability to write to an in-memory buffer, then insert it all at once into the persistent store. Implements it only in active_record, since it's meaningless in-memory and we don't have an optimized neo4j adapter yet.

Just for review right now, @CGDS wants to add bulk delete and bulk update capabilities so we'll work on that tomorrow. Also I haven't actually tested this against Postgres 9.5 yet, which is necessary to turn on some of the optimizations.


DMcKinnon-mdsol:
Should we be reviewing this?

HonoreDB:
I'll close it in favor of #63, which includes commits from this. Please do review that when you get a chance.

PR #63, opened 2016-07-28
Feature/batch spike

CGDS:
This PR, based off of #62, implements buffering at the adapter class level for writes of policy elements, assignments, associations, and deletes.  In the future, we may wish to limit buffering to a specific instance of an adapter for multithreaded usage of the policy machine, but all instances of the policy machine, should they wish to share a buffer, will also need to share the same instance of the policy machine storage adapter.

Regression results:
```
Finished in 37.19 seconds
728 examples, 0 failures, 12 pending
```

### DIFF
--- a/lib/policy_machine.rb
+++ b/lib/policy_machine.rb
@@ -1,3 +1,4 @@
+# coding: utf-8
 require 'policy_machine/policy_element'
 require 'policy_machine/association'
 require 'securerandom'
@@ -267,6 +268,10 @@ class PolicyMachine
     policy_machine_storage_adapter.transaction(&block)
   end
 
+  # TODO: For now, the adapter class itself mitigates buffering, so multiple instances of the
+  # adapter may share the same buffer.  In the future, we will move this to be adapter
+  # instance specific, so multiple policy machines sharing the same adapter may also
+  # share the same buffer without leaking global state.
   def bulk_persist
     adapter_class = policy_machine_storage_adapter.class
 
--- a/lib/policy_machine.rb
+++ b/lib/policy_machine.rb
@@ -269,15 +269,17 @@ class PolicyMachine
   end
 
   def bulk_persist
-    if policy_machine_storage_adapter.can_buffer?
+    adapter_class = policy_machine_storage_adapter.class
+
+    if adapter_class.can_buffer?
       begin
-        self.start_buffering!
+        adapter_class.start_buffering!
         result = yield
-        policy_machine_storage_adapter.bulk_persist!
+        adapter_class.persist_buffers!
         result
       ensure
-        self.stop_buffering!
-        policy_machine_storage_adapter.clear_buffers!
+        adapter_class.stop_buffering!
+        adapter_class.clear_buffers!
       end
     else
       yield
--- a/lib/policy_machine.rb
+++ b/lib/policy_machine.rb
@@ -241,7 +241,8 @@ class PolicyMachine
     define_method("create_#{pe_type}") do |unique_identifier, extra_attributes = {}|
       # when creating a policy element, we provide a unique_identifier, the uuid of this policy machine
       # and a policy machine storage adapter to allow us to persist the policy element.
-      pm_class.send(:create, unique_identifier, @uuid, @policy_machine_storage_adapter, extra_attributes)
+      meth = self.bulk_creating ? :create_later : :create
+      pm_class.send(meth, unique_identifier, @uuid, @policy_machine_storage_adapter, extra_attributes)
     end
 
     ##
@@ -265,6 +266,25 @@ class PolicyMachine
   # TODO: Possibly rescue NotImplementError and warn.
   def transaction(&block)
     policy_machine_storage_adapter.transaction(&block)
+  end
+
+  def bulk_create(&block)
+    if policy_machine_storage_adapter.respond_to?(:bulk_create!)
+      bulk_creator = clone
+      bulk_creator.bulk_creating = true
+      bulk_creator.instance_exec(block)
+      bulk_creator.bulk_create!
+    else
+      yield
+    end
+  end
+
+  protected
+
+  attr_accessor :bulk_creating
+
+  def bulk_create!
+    policy_machine_storage_adapter.bulk_create!
   end
 
   private
--- a/lib/policy_machine.rb
+++ b/lib/policy_machine.rb
@@ -242,8 +242,7 @@ class PolicyMachine
     define_method("create_#{pe_type}") do |unique_identifier, extra_attributes = {}|
       # when creating a policy element, we provide a unique_identifier, the uuid of this policy machine
       # and a policy machine storage adapter to allow us to persist the policy element.
-      meth = self.bulk_persisting ? :create_later : :create
-      pm_class.send(meth, unique_identifier, @uuid, @policy_machine_storage_adapter, extra_attributes)
+      pm_class.send(:create, unique_identifier, @uuid, @policy_machine_storage_adapter, extra_attributes)
     end
 
     ##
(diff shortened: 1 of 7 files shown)


lib/policy_machine.rb
```diff
@@ -1,3 +1,4 @@
+# coding: utf-8
```
HonoreDB: ?

lib/policy_machine_storage_adapters/active_record.rb
```diff
@@ -34,6 +34,52 @@ def self.load_db_adapter!
...
+    end
+
+    def buffers
+      self.class.buffers
+    end
+
+    def self.persist_buffers!
+      # TODO: if postgres needs all conflict targets specified, do we need this to be injectable?
+      # Also, we must move this into the pg adapter, since mysql behaves differently
+      # https://github.com/zdennis/activerecord-import/wiki/On-Duplicate-Key-Update
```
HonoreDB: I haven't tested it, but does mysql actually break here? I assumed passing it a truthy value would be enough and it'd just ignore the fact that it's an array.

lib/policy_machine_storage_adapters/active_record.rb
```diff
@@ -82,6 +128,75 @@ def self.serialize(store:, name:, serializer: nil)
...
+        #   to_create
+        # end
+      end
+
+      def self.assign_later(parent:, child:, buffer:)
+        buffer.merge!(parent => child)
+        :buffered
+      end
+
+      #TODO Are we not using PM UUID here?
```
HonoreDB: We're not, I kept in the method signature in case we want to change that as an optimization or something at some point, but right now the machine's implicitly identified by the other arguments and we're not bothering to validate.

lib/policy_machine_storage_adapters/active_record.rb
```diff
@@ -82,6 +128,75 @@ def self.serialize(store:, name:, serializer: nil)
...
+      def self.associate_later(user_attribute, operation_set, object_attribute, policy_machine_uuid, buffer)
+        buffer << [user_attribute, operation_set, object_attribute, policy_machine_uuid]
+      end
+
+      # NB: delete_all in AR bypasses relation logic, which shouldn't matter here.
+      def self.bulk_destroy(buffer)
+        id_groups = buffer.reduce(Hash.new { [] }) do |memo,(_,el)|
```
HonoreDB: Doesn't this need to be `Hash.new { |h,k| h[k]=[]}` ?

lib/policy_machine_storage_adapters/active_record.rb
```diff
@@ -82,6 +128,75 @@ def self.serialize(store:, name:, serializer: nil)
...
+        # else
+        #   to_create = new(attrs)
+
+        #   elements_to_create << to_create
+        #   to_create
+        # end
+      end
+
+      def self.assign_later(parent:, child:, buffer:)
+        buffer.merge!(parent => child)
```
HonoreDB: I feel like we can legitimately have multiple children for the same parent in the same buffer, which this looks like it doesn't support.

lib/policy_machine_storage_adapters/active_record.rb
```diff
@@ -82,6 +128,75 @@ def self.serialize(store:, name:, serializer: nil)
...
 
+      def buffers
+        pm_storage_adapter.buffers
+      end
+
+      def self.create_later(attrs, storage_adapter)
+        element = new(attrs)
+        storage_adapter.buffers[:upsert][element.unique_identifier] = element
+
+        # TODO: Why did we have to do this again? Don't get it
```
HonoreDB: Discussed this offline, this goes into part of a more general TODO to support non-upsertable databases like Postgres 9.4

lib/policy_machine.rb
CGDS: my bad, editor got away from me.

lib/policy_machine_storage_adapters/active_record.rb
CGDS: so I don't think it supports the same arg structure in the hash.  You might be right, but i don't really now.

lib/policy_machine_storage_adapters/active_record.rb
CGDS: I think I'm going to move your comment into my comment

lib/policy_machine_storage_adapters/active_record.rb
CGDS: ack, you are correct.  Overzealously trying to keep everything as a hash.

lib/policy_machine_storage_adapters/active_record.rb
CGDS: whoops, refactored my `+=` into a `<<`.  Which do you prefer?

HonoreDB: Ooh, I hadn't seen `+=` used like that, that's nifty. I guess I mildly prefer `<<` so you're not creating a throwaway single-element array.

lib/policy_machine_storage_adapters/active_record.rb
HonoreDB: [Looks to me like it actually honors the array](https://github.com/zdennis/activerecord-import/blob/3a29edae8a9ea576e350cc31b844320517510f02/test/support/mysql/import_examples.rb)

CGDS:
@DMcKinnon-mdsol || @mszenher, @HonoreDB and I have gone over this PR, ready for a third party final review / merge.

lib/policy_machine_storage_adapters/active_record.rb
```diff
@@ -129,6 +129,7 @@ def buffers
       def self.create_later(attrs, storage_adapter)
         element = new(attrs)
         storage_adapter.buffers[:upsert][element.unique_identifier] = element
+        element
```
HonoreDB: I'd be very surprised if this actually changes the return type of this method; unless it was changed recently the value of any assignment expression in Ruby was always the right-hand-side regardless of how complicated it was.

CGDS: it doesn't, just a stray commit when I was seeing some odd behavior in dalton. will reverse.

ayau-mdsol:
@DMcKinnon-mdsol, @mszenher Can one of you work on getting this merged as soon as possible? It is needed for us to merge the dalton changes to optimize performance for grants. Thanks.

lib/policy_machine_storage_adapters/active_record.rb
```diff
@@ -82,6 +121,52 @@ def self.serialize(store:, name:, serializer: nil)
...
+      def self.bulk_destroy(elements)
+        id_groups = elements.reduce(Hash.new { |h,k| h[k] = [] }) do |memo,(_,el)|
+          if el.is_a?(UserAttribute) || el.is_a?(ObjectAttribute)
+            memo[el.class] << el.id
+          end
+
+          memo
+        end
+
+        PolicyElement.where(unique_identifier: elements.keys).delete_all
```
DMcKinnon-mdsol: @CGDS @HonoreDB Don't these `delete_all`s mean that we won't get callbacks for Archon notifications or audits? Don't we want those?

HonoreDB: In our consuming app, which shall remain nameless, the after_delete callbacks aren't on these ActiveRecord objects but on ActiveModel objects wrapping them, so they still get hit. The callbacks we skip by using `delete_all` are just deleting dependent objects, which we handle manually in the next three lines.

DMcKinnon-mdsol: 👍

PR #65, opened 2016-08-08
Fix/bulk json

CGDS:
Bulk Insert does not handle 'special' AR insert types (hstore, jsonb, etc) if the AR magic to use attributes delegated to the store is the method used to set them.  This PR escapes that problem by reassigning all attributes at the top level before persistence.  @DMcKinnon-mdsol and @HonoreDB, please review

### DIFF
--- a/lib/policy_machine_storage_adapters/active_record.rb
+++ b/lib/policy_machine_storage_adapters/active_record.rb
@@ -70,9 +70,9 @@ module PolicyMachineStorageAdapter
       # during import, we set all attributes again in bulk here.  It is important
       # that these changes are mutative, since the default ActiveRecord magic
       # being relied on for assignments and associations will break, otherwise
-      buffers[:upsert].values.each { |element| element.attributes = element.attributes.slice(*column_keys) }
+      buffers[:upsert].values.each { |el| el.attributes = el.attributes.slice(*column_keys) }
 
-      PolicyElement.import(elements_to_upsert, on_duplicate_key_update: column_keys.map(&:to_sym) - [:id])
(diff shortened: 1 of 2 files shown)


lib/policy_machine_storage_adapters/active_record.rb
```diff
@@ -64,8 +64,18 @@ def buffers
...
     def self.persist_buffers!
-      column_keys = PolicyElement.column_names.map(&:to_sym)
-      PolicyElement.import(buffers[:upsert].values, on_duplicate_key_update: column_keys - [:id])
+      column_keys = PolicyElement.column_names
+
+      # Because activerecord-import cannot yet handle arbitrary serialized values
+      # during import, we set all attributes again in bulk here.  It is important
+      # that these changes are mutative, since the default ActiveRecord magic
+      # being relied on for assignments and associations will break, otherwise
+      elements_to_upsert = buffers[:upsert].values.map do |element|
```
HonoreDB: `buffers[:upsert].values.each { |element| element.attributes = element.attributes.slice(*column_keys)` seems clearer to me, makes it more obvious that it mutates. You could also definite an instance method on PolicyElement for this and make it
`values.each(&:prep_for_import!)`

HonoreDB:
LGTM

lib/policy_machine_storage_adapters/active_record.rb
CGDS: I'd prefer not to permanentize it like that, since hopefully this will be addressed at some point relatively soon with active record import.  but the each thing is a good call.

CGDS:
```
Finished in 46.58 seconds
764 examples, 0 failures, 12 pending
Coverage report generated for RSpec 747 / 832 LOC (89.78%) covered.
```

PR #66, opened 2016-08-25
Join on id, not on implicit cast of unique_identifier string

HonoreDB:
Specs pass. The bulk cleanup of orphaned assignments got wonky and was referencing the unique_identifier, not the id, of deleted elements. If the unique_identifier happened to be a string starting with a number, this would be cast to that number and would delete random assignments.

### DIFF
--- a/lib/policy_machine_storage_adapters/active_record.rb
+++ b/lib/policy_machine_storage_adapters/active_record.rb
@@ -149,7 +149,8 @@ module PolicyMachineStorageAdapter
         end
 
         PolicyElement.where(unique_identifier: elements.keys).delete_all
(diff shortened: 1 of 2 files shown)


lib/policy_machine_storage_adapters/active_record.rb
```diff
@@ -149,7 +149,8 @@ def self.bulk_destroy(elements)
         end
 
         PolicyElement.where(unique_identifier: elements.keys).delete_all
-        Assignment.where(parent_id: elements.keys).delete_all
```
CGDS: yeesh.  Thanks @HonoreDB

CGDS:
lgtm.  merging in preparation for tagging

PR #68, opened 2016-09-06
Admit that we're preloading objects into memory and just fetch their ids instead

HonoreDB:
Specs pass, seems like it shaves about a second off of the long tail we're trying to fix.


DMcKinnon-mdsol:
👍

HonoreDB:
Got rid of the pointless `|` between heterogeneously typed arrays.

CGDS:
cool, looks good @HonoreDB.  Assume specs still pass?

HonoreDB:
Specs still pass.

DMcKinnon-mdsol:
👍

PR #69, opened 2016-09-06
Exists generates better sql than any

HonoreDB:
Slight improvement (90 ms) on the test case. Specs pass.


DMcKinnon-mdsol:
Just for some additional info, @HonoreDB explained to me irl that `any?` uses something like count(*) > 0, and exists? uses limit 1, which is cooler.

CGDS:
I like cooler.

PR #72, opened 2016-09-14
Remove = from the attribute key

HonoreDB:
As promised, a fix using psychic powers since we can't repro this locally. What I think was going down: we hit method_missing when trying to set a new attribute having never previously referenced it, and the affected code erroneously treats "foo=" as the attribute name. Then, another code path hits `store_attributes` again, and that key becomes a methodized attribute name. Then we try to set `foo=` again, and this time it's defined as an arity-0 getter instead of an arity 1 setter.

Specs pass.


DMcKinnon-mdsol:
@HonoreDB you said this fixed it on sandbox, yes? If so, I'll merge now :)

HonoreDB:
Yup

PR #79, opened 2016-10-13
Fix abstraction layer skip in batch find

HonoreDB:
The regular finds go through the PolicyElement classes, which wrap the results of the underlying storage engine. The batch finds bypass this to go right to the storage adapter. This usually doesn't have much effect due to duck typing, but means we miss some methods only provided in the wrapper. This PR puts the wrapper back.

### DIFF
--- a/lib/policy_machine.rb
+++ b/lib/policy_machine.rb
@@ -197,8 +197,12 @@ class PolicyMachine
   ##
   # Search for and iterate over a collection in batches
   def batch_find(type:, query: {}, config: {}, &blk)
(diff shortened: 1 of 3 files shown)


DMcKinnon-mdsol:
@HonoreDB Mind regressing specs locally and posting the results here?

Any comments @CGDS or @PoloPro ?

HonoreDB:
My specs pass in `1 minute 59.66 seconds` until they hit the neo4j suite, which seems to've been broken by an earlier PR.

DMcKinnon-mdsol:
Sigh. That's sad, but non-blocking. Merging in 30 minutes unless objections.

PR #81, opened 2016-10-24
[MCC-250997] Disable merge join on psql nested queries

DMcKinnon-mdsol:
This PR prevents psql from using merge join on nested queries.

Opening a PR to Dalton to check its build (on the release branch) using this commit.

All specs pass locally:
Finished in 54.39 seconds
769 examples, 0 failures, 12 pending

@mdsol/team-04 please review.


DMcKinnon-mdsol:
Pending this build passing: https://github.com/mdsol/dalton/pull/1588

HonoreDB:
I don't think this will reliably work, it looks like it'd go:

```
BEGIN TRANSACTION
set local enable_mergejoin = false
<generate an activerecord scope in-memory>
COMMIT
run query
```

So the local enable_mergejoin variable is applied to the query only if we're in a larger transaction. You might have to do some fancy dependency inversion stuff between the AR adapter and the Postgres adapter to get the ordering right.

lib/query_plan_hints.rb
```diff
@@ -0,0 +1,17 @@
+module QueryPlanHints
+
+  # Method decorator to ensure not using a specific query plan
+  # during postgres privilege deriviations
+  def disable_mergejoin(method_name)
+    if ActiveRecord::Base.connection.is_a? ActiveRecord::ConnectionAdapters::PostgreSQLAdapter
```
HonoreDB: Is there a way to use `PolicyMachineStorageAdapter::ActiveRecord::Assignment` instead of `ActiveRecord::Base` here and below? Not sure the initialization timing works out of the box, but it'd play better with projects that use multiple database connections.

DMcKinnon-mdsol:
Updated to use a block. All specs pass:
Finished in 52.85 seconds
769 examples, 0 failures, 12 pending

HonoreDB:
Looks good to me, may I merge?

DMcKinnon-mdsol:
Please do!

PR #89, opened 2017-03-02
Don't spam the same warning repeatedly within a single process

HonoreDB:
While we're here. I'm not taking out the warning entirely, because it's legitimate (old man shakes fist at clouds), but no reason to fill up sumo and rspec logs with it.


CGDS:
I applaud this sentiment.

DMcKinnon-mdsol:
My heart beat still.

DMcKinnon-mdsol:
Merging in 15 unless objections

PR #92, opened 2017-03-09
Remove some db calls

DMcKinnon-mdsol:
This PR prevents us from making a bunch or short but unnecessary calls to the db on buffered stuff.

@mdsol/team-04 please review.


PoloPro:
LGTM, unless you want to add tests to confirm the db isn't being hit in the empty case.

DMcKinnon-mdsol:
I don't think that's necessary, but I'll certainly add the tests if @HonoreDB or @CGDS wants them.

HonoreDB:
Nah, I'm good.

PoloPro:
Alright, will merge in 15m unless @CGDS has concerns

PR #99, opened 2017-03-21
Pluck virtual attributes

PoloPro:
A brief implementation of `pluck_all_in_batches` supporting virtual attribute queries. 

Current method is to 
1. Intercept plucks with `fields` arguments that are extra_attributes
2. Remove the offending fields and replace them with `:extra_attributes`
3. Perform the `ActiveRecord::Relation.select` as normal
4. ???

The return value, if it contains _any_ extra_attributes, will contain _all_ of the extra_attributes. Alternately, there could be a step 4 which reverses the shell game and substitutes the original fields back into the return value.

Opened for discussion and alternate methodology. @mdsol/team04

### DIFF
--- a/lib/policy_machine_storage_adapters/active_record.rb
+++ b/lib/policy_machine_storage_adapters/active_record.rb
@@ -375,10 +375,10 @@ module PolicyMachineStorageAdapter
 
         # Fields must include a primary key to avoid ActiveRecord errors
         fields << :id
-        if extras = extra_attribute_fields(all.first, fields)
-          fields = fields - extras
-          fields << :extra_attributes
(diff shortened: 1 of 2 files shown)


lib/policy_machine_storage_adapters/active_record.rb
```diff
@@ -370,9 +370,16 @@ class OperationsPolicyElementAssociation < ::ActiveRecord::Base
       end
 
       define_method("pluck_all_of_type_#{pe_type}") do |fields:, options: {}|
+        all = method("find_all_of_type_#{pe_type}").call(options)
+        return unless all.present?
```
DMcKinnon-mdsol: Shouldn't we return an empty array.

PR #104, opened 2017-03-31
[MCC-291327] Retry find_or_create on index uniqueness violations

DMcKinnon-mdsol:
@jlo188 Noticed that we're receiving quite a few `PG::UniqueViolation` errors in all environments. I traced this back to [AR's find_or_create method not being atomic](http://api.rubyonrails.org/classes/ActiveRecord/Relation.html#method-i-find_or_create_by). 

@mdsol/team-04 please review.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,7 @@
 # Changelog
 
 ## 1.4.2
-* Fix the ActiveRecord Adapter to retry on `find_or_create` uniqueness violations.
+* Fix the ActiveRecord Adapter to use upserts instead of find or creates.
 
 ## 1.4.1
 * Standardized the return value of batch_pluck.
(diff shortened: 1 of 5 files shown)


HonoreDB:
Mildly scary pattern; if there's a unique constraint that's less specific than the `first_or_create` query, this is an infinite loop (and actually infinite, it won't throw a stack level too deep error, so it'll tie up a thread and database connection). I'd feel more comfortable with, instead of `retry`, a `find(same_params) || raise "Something weird happened"`.

HonoreDB:
Also don't you need to do a manual reconnect after hitting a PG error inside a transaction? Or is ActiveRecord smarter than that?

HonoreDB:
You'd probably get faster and safer using `import` with the `on_duplicate_key_ignore` option, which we already do here elsewhere. No postgres error, no extra read.

PoloPro:
I don't know the quality of this pattern, but you could also have a `max_retries`. Would prevent explosions.

HonoreDB:
LGTM

PR #106, opened 2017-04-24
Fix/ruby 24 specs

HonoreDB:
The Policy Machine does type checking. A lot.

Ruby 2.4 changes the type of numbers from `Fixnum/Bignum < Integer` to just `Integer`.

That change breaks assertions in specs testing the error message returned by typechecking.

This PR fixes those specs.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,6 @@
 language: ruby
 rvm:
+  - 2.4.0
   - 2.2.3
 env:
   global:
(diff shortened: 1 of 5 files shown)


DMcKinnon-mdsol:
Thanks Aaron!

PoloPro:
Looks good. Nice catch Aaron.

DMcKinnon-mdsol:
Mind also updating the travis.yml to also regress specs with ruby 2.4.0?

PR #110, opened 2017-04-27
[MCC-276230] Fix/new import code

CGDS:
So #92 added some import code, but the problem with AR import on duplicate key ignore is that it doesn't update the id of the passed record, so subsequent `operations=` method would fail to create an operation policy element association record.  Also added the partial uniqueness constraint for the policy element association iff it has a operation_set_id here.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,7 @@
 # Changelog
 
 ## 1.5.1
-* Fix a bug in some of the new import code that prevented operation policy element
+* Fix a bug in some of the new import code that prevented operation policy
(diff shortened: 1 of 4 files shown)


CHANGELOG.md
```diff
@@ -1,5 +1,10 @@
 # Changelog
 
+## 1.5.1
+* Fix a bug in some of the new import code that prevented operation policy element
+  element associations from saving correctly when encountering duplicates, also
```
DMcKinnon-mdsol: This message has `element` twice.

CGDS: its `elemental`

PR #112, opened 2017-05-02
Upgrade to RSpec v3.5

PoloPro:
A long overdue upgrade to the current stable version of RSpec from v2.13. Enables more powerful comparators and the cleaner `expect` syntax for specs.

Also updates all `should` matchers to `expects` and utilizes the `contain_exactly` matcher where applicable.

### DIFF
--- a/spec/policy_machine_storage_adapters/active_record_spec.rb
+++ b/spec/policy_machine_storage_adapters/active_record_spec.rb
@@ -28,27 +28,27 @@ describe 'ActiveRecord' do
     describe 'find_all_of_type' do
 
       it 'warns once when filtering on an extra attribute' do
-        Warn.should_receive(:warn).once
+        expect(Warn).to receive(:warn).once
         2.times do
-          policy_machine_storage_adapter.find_all_of_type_user(foo: 'bar').should be_empty
+          expect(policy_machine_storage_adapter.find_all_of_type_user(foo: 'bar')).to be_empty
         end
       end
 
       context 'an extra attribute column has been added to the database' do
 
         it 'does not warn' do
-          Warn.should_not_receive(:warn)
-          policy_machine_storage_adapter.find_all_of_type_user(color: 'red').should be_empty
+          expect(Warn).to_not receive(:warn)
+          expect(policy_machine_storage_adapter.find_all_of_type_user(color: 'red')).to be_empty
         end
 
         it 'only returns elements that match the hash' do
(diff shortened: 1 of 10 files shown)


DMcKinnon-mdsol:
Nice! Thanks @PoloPro.

spec/policy_machine_storage_adapters/active_record_spec.rb
```diff
@@ -181,9 +181,9 @@
               caffeinated.unassign(decaffeinated)
             end
 
-            expect(user.connected?(decaffeinated)).to be true
-            expect(user.connected?(caffeinated)).to be true
-            expect(caffeinated.connected?(decaffeinated)).to be false
+            expect(user.connected?(decaffeinated)).to be_truthy
```
DMcKinnon-mdsol: `be true` and `be_truthy` aren't the same. I think it's only suggested that methods ending in `?` return truthy or falsey values (as opposed to strictly `true` or `false`), but given that we're already testing for that level of detail, why not make these `eq true`?

DMcKinnon-mdsol: Did you use a tool for this upgrade or do it by hand?

PoloPro: By hand. And no, those aren't the same. All cases of `be_true` should now be `be_truthy`, `be true` should be unchanged.

PoloPro: Looks like a find-replace went one step too far. There aren't too many instances of that switch , I'll ensure they all retain meaning.

DMcKinnon-mdsol:
Mind also updating the `version.rb` file?

DMcKinnon-mdsol:
Also, for future reference, I think we've used https://github.com/yujinakayama/transpec with some success in the past.

spec/policy_machine_storage_adapters/active_record_spec.rb
```diff
@@ -384,6 +384,8 @@
       @pm2_op = @pm2.create_operation('pm2 op')
 
       @user_attributes = (1..n).map { |i| @pm.create_user_attribute("ua#{i}") }
+      @user_attributes.first.update(color: 'red')
```
DMcKinnon-mdsol: Why was this change necessary?

PoloPro: It's absolutely imperative that the first user attribute be red.

PR #113, opened 2017-05-02
Added optional filtering to descendants and ancestors methods

PoloPro:
This PR adds optional attribute filtering to the ActiveRecord `descendants` and `ancestors` methods. This functions similarly to the `query` optional argument in the find methods, and helps limit memory overhead when a smaller subset of results is needed.

Also updates the `method_missing` in the core PolicyElement class to pass arguments through, rather than stripping them.

Finished in 1 minute 38.03 seconds (files took 0.90146 seconds to load)
1553 examples, 0 failures, 19 pending

### DIFF
--- a/lib/policy_machine_storage_adapters/active_record.rb
+++ b/lib/policy_machine_storage_adapters/active_record.rb
@@ -137,7 +137,15 @@ module PolicyMachineStorageAdapter
       end
 
       def descendants(filters = {})
-        Assignment.descendants_of(self).where(filters)
+        if filters.empty?
+          Assignment.descendants_of(self)
+        else
+          valid_filters = PolicyElement.columns_hash.map { |column| column.first.to_sym }
+          filters.each_key do |key|
+            raise(ArgumentError, "#{key} is not a valid filter") unless valid_filters.include?(key)
+          end
+          Assignment.descendants_of(self).where(filters)
+        end
       end
 
       def link_descendants
@@ -145,7 +153,15 @@ module PolicyMachineStorageAdapter
       end
 
       def ancestors(filters = {})
-        Assignment.ancestors_of(self).where(filters)
+        if filters.empty?
+          Assignment.ancestors_of(self)
+        else
+          valid_filters = PolicyElement.columns_hash.map { |column| column.first.to_sym }
+          filters.each_key do |key|
+            raise(ArgumentError, "#{key} is not a valid filter") unless valid_filters.include?(key)
+          end
+          Assignment.ancestors_of(self).where(filters)
+        end
       end
 
       def link_ancestors
--- a/lib/policy_machine_storage_adapters/active_record.rb
+++ b/lib/policy_machine_storage_adapters/active_record.rb
@@ -137,27 +137,17 @@ module PolicyMachineStorageAdapter
       end
 
       def descendants(filters = {})
-        if filters.present?
-          valid_filters = PolicyElement.column_names
-          filters.stringify_keys.each_key do |key|
-            raise(ArgumentError, "#{key} is not a valid filter") unless valid_filters.include?(key)
-          end
-        end
+        assert_valid_filters!(filters)
         Assignment.descendants_of(self).where(filters)
+      end
+
+      def ancestors(filters = {})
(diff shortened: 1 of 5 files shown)


DMcKinnon-mdsol:
Thanks @PoloPro! This lgtm. Out of an abundance of paranoia wouldn't mind also seeing tests for when a filter hits none/all descendants/ancestors, or when the filters key is not a valid attribute.

lib/policy_machine_storage_adapters/active_record.rb
```diff
@@ -137,15 +137,31 @@ def store_attributes
...
+        end
       end
 
       def link_descendants
         LogicalLink.descendants_of(self)
       end
 
       def ancestors(filters = {})
-        Assignment.ancestors_of(self).where(filters)
+        if filters.empty?
```
DMcKinnon-mdsol: Is this check even necessary? I would've thought that this holds:
```
Assignment.ancestors_of(self).where({}) == Assignment.ancestors_of(self)
```

lib/policy_machine_storage_adapters/active_record.rb
```diff
@@ -137,15 +137,31 @@ def store_attributes
...
+          Assignment.ancestors_of(self)
+        else
+          valid_filters = PolicyElement.columns_hash.map { |column| column.first.to_sym }
```
DMcKinnon-mdsol: `PolicyElement.column_names` is a thing.

lib/policy_machine_storage_adapters/active_record.rb
PoloPro: Yes, but I'd like to remove even the empty `where` call in the base case, for whatever minor performance gain that is. It's core enough of a method that I don't want to clutter it with chaining unless necessary.

lib/policy_machine_storage_adapters/active_record.rb
DMcKinnon-mdsol: Seems a bit fragile to not account for string keys in case someone wants to use hash rocket syntax. Might be worth throwing a `with_indifferent_access` in somewhere.

PoloPro: Glorious, will add.

lib/policy_machine_storage_adapters/active_record.rb
DMcKinnon-mdsol: I'd expected it to handle that case and be just as performant. I think the point of it accepting empty params is that we don't have to write if clauses like this one.

PoloPro: I thought empty params were because Ruby doesn't support method overloading.

In either case, I checked the `where` implementation and it just does the same check on its end (for empty args), and returns self. I'm fine removing the check here and delegating

spec/policy_machine_storage_adapters/active_record_spec.rb
```diff
@@ -400,9 +401,44 @@
...
+        end
+
+        it 'returns appropriate results when filters apply to no descendants' do
+          expect(@u1.descendants(color: 'taupe')).to be_empty
+          expect { @u1.descendants(not_a_real_attribute: 'fake') }.to raise_error(ArgumentError)
+        end
+
+        it 'returns appropriate results when filters apply to all descendants' do
+          all_descendants = @user_attributes.map(&:stored_pe) + [@new_ua.stored_pe]
+          expect(@u1.descendants({})).to match_array(all_descendants)
```
PoloPro: `contain_exactly` takes a comma-separated list of elements, `match_array` takes an array. It's incredibly minor, just depends on what data you have lying around.

DMcKinnon-mdsol:
Re:
> It would be nice if this also covered the in memory adapter and the specs could be moved to shared examples, but I get it if you think that is out of scope.

Are we still waiting for those, or are they going in a separate PR?

PoloPro:
The in memory adapter doesn't implement `ancestors` or `descendants`. Connor and I talked offline, and the only adapter that could use the update would be the mysql, which we're planning to deprecate anyway.

PR #116, opened 2017-05-09
Add optional filters to parents and children attribute accessors

PoloPro:
In addition to the filters added to the `descendants` and `ancestors` methods, the `parents` and `children` attributes also need that functionality. 

Methodology is the same, adding `where` clauses to the ActiveRecord relations via chaining. Tests and version updated to match.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+## 1.5.3
+* Add optional filtering to parents and children attribute accessors in the ActiveRecord storage
+  adapter.
+
 ## 1.5.2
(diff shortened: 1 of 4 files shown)


DMcKinnon-mdsol:
Should we also add this feature to links while we're in here?

PoloPro:
PR now implements filters for all ActiveRecord associations. Associations themselves have been renamed to avoid Rails automagic caching, which fails when delegating to the auto-generated accessor. Specs added for all association filters.

DMcKinnon-mdsol:
Merging in 30 unless objections.

PR #119, opened 2017-06-06
[MCC-305829] Give precedence to column accessors over extra_attributes

PoloPro:
`store_attributes` previously would overwrite accessors for actual columns with extra_attributes accessors when there was a conflict. This PR gives precedence to the column accessors, only defining accessors for non-existent attributes.

### DIFF
--- a/lib/policy_machine_storage_adapters/active_record.rb
+++ b/lib/policy_machine_storage_adapters/active_record.rb
@@ -132,8 +132,10 @@ module PolicyMachineStorageAdapter
 
       # Uses ActiveRecord's store method to methodize new attribute keys in extra_attributes
       def store_attributes
+        # Do not overwrite accessors for existing columns
         @extra_attributes_hash = extra_attributes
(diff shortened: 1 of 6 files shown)


HonoreDB:
LGTM

lib/policy_machine_storage_adapters/active_record.rb
```diff
@@ -132,8 +132,10 @@ def respond_to_missing?(meth, *args)
 
       # Uses ActiveRecord's store method to methodize new attribute keys in extra_attributes
       def store_attributes
+        # Do not overwrite accessors for existing columns
         @extra_attributes_hash = extra_attributes
-        self.class.store_accessor(:extra_attributes, @extra_attributes_hash.keys)
+        column_attributes = PolicyElement.column_names.map(&:to_sym)
+        self.class.store_accessor(:extra_attributes, @extra_attributes_hash.except(column_attributes).keys)
```
fpcyan: should there be a validation or something that extra_attributes doesn't have column attributes in it in the first place?

PR #120, opened 2017-07-17
pluck_parents and pluck_children

PoloPro:
This PR adds `pluck_parents` and `pluck_children` methods to the ActiveRecord storage adapter. These filtered graph traversals only return specified fields from the records, eliminating potentially unneeded overhead.

This will increment the version of the PolicyMachine to 1.6.0, reflecting new functionality.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,8 @@
 # Changelog
 
+## 1.6.0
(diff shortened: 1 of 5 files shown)


PoloPro:
Labeling as Do Not Merge until it can be verified on the Dalton side.

PoloPro:
Closing for now, as it doesn't appear necessary to solve current bottlenecks.

PR #121, opened 2017-08-14
[MCC-321135] Remove the OperationsPolicyElementsAssociation dependency in favor of assignments between Operations Sets and Operations

DMcKinnon-mdsol:
This PR updates the gem to rely on the assignments table using recursive queries between operations and operation sets instead of the operations policy elements associations table. Preliminary performance tests on sandbox indicate that this method _may_ be 0% - 20% slower for some privilege derivations.

~~An n+1 query has been added to the `accessible_objects` method in the ActiveRecord adapter (In Dalton, this is used for [bulk privilege gets](https://github.com/mdsol/dalton/blob/develop/app/models/privilege.rb#L111-L148), the less frequently used privilege index). While I'm sure some SQL-foo exists to do this in a more efficient way, I haven't gotten anything working yet. I'll keep at it, though. Suggestions welcome.~~

@mdsol/team04 please review.


CHANGELOG.md
```diff
@@ -1,7 +1,8 @@
 # Changelog
 
-## 1.5.4
```
CGDS: why remove the 1.5.4 section?

DMcKinnon-mdsol: Because we never tagged a 1.5.4 gem.

lib/migrations/add_initial_policy_machine_tables.rb
```diff
@@ -31,11 +31,5 @@ def change
     add_index :assignments, [:parent_id, :child_id], unique: true
     add_index :assignments, [:child_id]
 
-    create_table :operations_policy_element_associations, id: false do |t|
```
CGDS: any other changes needed for the generator?

lib/policy_machine.rb
```diff
@@ -127,7 +126,9 @@ def is_privilege_ignoring_prohibitions?(user_or_attribute, operation, object_or_
       raise(ArgumentError, "options[:associations] cannot be empty") if associations.empty?
       raise(ArgumentError, "expected each element of options[:associations] to be a PM::Association") unless associations.all?{|a| a.is_a?(PM::Association)}
 
-      associations.keep_if{ |assoc| assoc.includes_operation?(operation) }
+      associations.keep_if do |association|
+        association.operation_set.connected?(operation)
```
CGDS: is this an n+1?  or actually, if operation_sets also aren't being loaded 2n + 1 i guess?

lib/migrations/add_initial_policy_machine_tables.rb
DMcKinnon-mdsol: What did you have in mind?

lib/policy_machine/association.rb
```diff
@@ -35,34 +24,15 @@ def initialize(stored_user_attribute, stored_set_of_operation_objects, stored_op
...
       raise(ArgumentError, "user_attribute_pe must be a UserAttribute.") unless user_attribute_pe.is_a?(PM::UserAttribute)
       unless user_attribute_pe.policy_machine_uuid == policy_machine_uuid
         raise(ArgumentError, "user_attribute_pe must be in policy machine with uuid #{policy_machine_uuid}")
       end
 
-      # argument errors for set_of_operation_objects
-      raise(ArgumentError, "set_of_operation_objects must be a Set of Operations") unless set_of_operation_objects.is_a?(Set)
-      raise(ArgumentError, "set_of_operation_objects must not be empty") if set_of_operation_objects.empty?
-      set_of_operation_objects.each do |op|
-        unless op.is_a?(PM::Operation)
```
CGDS: do you think we need / do we have equivalent type checking on operation sets?

lib/policy_machine_storage_adapters/active_record.rb
```diff
@@ -140,12 +140,12 @@ def store_attributes
 
       def descendants(filters = {})
         assert_valid_filters!(filters)
-        Assignment.descendants_of(self).where(filters)
+        Assignment.descendants_of(self.id).where(filters)
```
CGDS: why did we change this method signature?

lib/policy_machine_storage_adapters/active_record.rb
```diff
@@ -244,26 +244,17 @@ def self.bulk_link(pairs_hash)
...
-                                        object_attribute_id: object_attribute.id,
-                                        operation_set_id: operation_set.id),
-           set_of_operation_objects]
+        associations.map! do |user_attribute, operation_set, object_attribute, policy_machine_uuid|
+          association =
+            PolicyElementAssociation.new(
+              user_attribute_id: user_attribute.id,
+              object_attribute_id: object_attribute.id,
+              operation_set_id: operation_set.id)
+          [association, operation_set]
```
CGDS: do we actually need to return the operation set now?

lib/policy_machine_storage_adapters/active_record.rb
```diff
@@ -290,11 +281,14 @@ class Object < ObjectAttribute
     end
 
     class Operation < PolicyElement
-      has_and_belongs_to_many :policy_element_associations, class_name: 'PolicyMachineStorageAdapter::ActiveRecord::PolicyElementAssociation', join_table: 'operations_policy_element_associations'
```
CGDS: we could add this scope back using a recursive through

lib/policy_machine_storage_adapters/active_record.rb
```diff
@@ -663,22 +636,32 @@ def transaction(&block)
...
-        scoped_privileges_single_policy_class(user_or_attribute, object_or_attribute, options)
-      else
-        scoped_privileges_multiple_policy_classes(user_or_attribute, object_or_attribute, policy_classes_containing_object, options)
-      end
+
+      operations =
+        if policy_classes_containing_object.count < 2
+          accessible_operations(user_or_attribute.id, object_or_attribute.id)
+        else
+          policy_classes_containing_object.flat_map do |policy_class|
```
CGDS: this seems like a significant change, is this just basically unravelling the logic that was previously in `scoped_privileges_multiple_policy_classes`?

lib/policy_machine_storage_adapters/active_record.rb
```diff
@@ -716,67 +712,29 @@ def accessible_objects(user_or_attribute, operation, options = {})
...
-          if operation.is_a?(class_for_type('operation'))
-            associations_between(user_or_attribute, object_or_attribute).where(id: operation.policy_element_associations.to_a, object_attribute_id: pc.ancestors).any?
-          else
-            associations_between(user_or_attribute, object_or_attribute).joins(:operations).where(policy_elements: {unique_identifier: operation}, object_attribute_id: pc.ancestors).any?
-          end
-        end
-      end
-    end
+        associations =
+          PolicyElementAssociation.where(
```
CGDS: could we filter this down farther by filtering this down with relevant opset ids when provided?  I guess at that point all we'd want is a count greater than one, no?  seems like that would be a nice perf boost with a limit clause for very authorized users

lib/policy_machine_storage_adapters/active_record/postgresql.rb
```diff
@@ -9,38 +9,66 @@ class Assignment < ::ActiveRecord::Base
       belongs_to :child, class_name: 'PolicyElement', foreign_key: :child_id
 
       def self.transitive_closure?(ancestor, descendant)
-        descendants_of(ancestor).include?(descendant)
+        descendants_of(ancestor.id).include?(descendant)
       end
 
-      def self.descendants_of(element_or_scope)
+      def self.descendants_of(ids)
```
CGDS: im not sure i like going to ids, seems like it would make it harder to collapse this into a single query compared to a scope without having to touch all the calling sights in future

lib/policy_machine_storage_adapters/active_record/postgresql.rb
```diff
@@ -9,38 +9,66 @@ class Assignment < ::ActiveRecord::Base
...
+          PolicyElement.where('"policy_elements"."id" IN (SELECT assignments__recursive.parent_id FROM (WITH RECURSIVE "assignments__recursive" AS ( SELECT "assignments"."id", "assignments"."parent_id", "assignments"."child_id" FROM "assignments" WHERE "assignments"."child_id" in (?) UNION SELECT "assignments"."id", "assignments"."parent_id", "assignments"."child_id" FROM "assignments" INNER JOIN "assignments__recursive" ON "assignments__recursive"."parent_id" = "assignments"."child_id" ) SELECT "assignments__recursive".* FROM "assignments__recursive") AS "assignments__recursive")', ids)
         end
       end
 
+      # Returns the operation set IDs from the given list where the operation is
+      # a descendant of the operation set.
+      # TODO: Generalize this so that we can arbitrarily filter recursive assignments calls.
+      def self.filter_operation_set_list_by_assigned_operation(operation_set_ids, operation_id)
+        query =
+          "WITH RECURSIVE assignments_recursive AS (
```
CGDS: mild preference for heredoc

lib/policy_machine_storage_adapters/active_record/postgresql.rb
```diff
@@ -61,9 +89,9 @@ def self.descendants_of(element_or_scope)
...
+          PolicyElement.where('"policy_elements"."id" IN (SELECT logical_links__recursive.link_child_id FROM (WITH RECURSIVE "logical_links__recursive" AS ( SELECT "logical_links"."id", "logical_links"."link_child_id", "logical_links"."link_parent_id" FROM "logical_links" WHERE "logical_links"."link_parent_id" = ? UNION SELECT "logical_links"."id", "logical_links"."link_child_id", "logical_links"."link_parent_id" FROM "logical_links" INNER JOIN "logical_links__recursive" ON "logical_links__recursive"."link_child_id" = "logical_links"."link_parent_id" ) SELECT "logical_links__recursive".* FROM "logical_links__recursive") AS "logical_links__recursive")', element_or_scope.first.id)
```
CGDS: im having a hard time seeing what changed in these queries, is just the UNION ALL -> UNION ?  isn't that going to be a perf hit?  or is it negligible in practice?

DMcKinnon-mdsol:
Build is fixed. Travis changed this repo to use docker builds, and for some reason, that broke the gem load paths for just ruby 2.4.0.

lib/policy_machine_storage_adapters/active_record/postgresql.rb
DMcKinnon-mdsol: Good catch. Yea, I just changed the "UNION ALLs" to "UNIONs" hoping to get a perf boost by removing duplicate rows on successive "recursions" (psql does fake recursion), but after some research, you're definitely right that calling distinct on every iteration will incur a perf hit. I'll change it back.

lib/policy_machine.rb
DMcKinnon-mdsol: Yes, it is. It's also my understanding that only the in memory adapter uses this, so I didn't think twice about it.

lib/policy_machine/association.rb
DMcKinnon-mdsol: You just mean something that checks that the operation set is an operation set? Yea, that's directly below this...you added it 😝 .

lib/policy_machine_storage_adapters/active_record.rb
DMcKinnon-mdsol: Nope! Good call. Removed.

lib/policy_machine_storage_adapters/active_record.rb
DMcKinnon-mdsol: Yup! I _think_ it does the same thing in much less code.

lib/policy_machine_storage_adapters/active_record/postgresql.rb
DMcKinnon-mdsol: It would. It also removed a join because it's currently not collapsed. But sure we can hash that out on a separate PR. Changed it back for now.

lib/policy_machine_storage_adapters/active_record.rb
DMcKinnon-mdsol: In what context would we provide "relevant opset ids"?

lib/policy_machine_storage_adapters/active_record.rb
DMcKinnon-mdsol: Do we still need this scope? Which code should it replace?

DMcKinnon-mdsol: How does a recursive through look? I did some poking around on the interwebs and didn't stumble across an example.

lib/migrations/add_initial_policy_machine_tables.rb
DMcKinnon-mdsol: Like adding a generator to drop the operations policy elements association table? I'm fine just doing that in a Dalton migration if you are.

DMcKinnon-mdsol:
@mdsol/team04 please review when you get a chance. Ask questions if you have them. I can also sit down with people if you're not sure what's going on here or why.

lib/policy_machine.rb
DMcKinnon-mdsol: Argh, it looks like this could use the AR storage adapter if associations are provided to the privilege call. That's not a thing we do, but it's a thing we _could_ do. I love the policy machine. I'll revisit this.

fpcyan:
> bulk privilege gets, the less frequently used privilege index

I challenge that bulk privilege get is use exceedingly frequently and is, in fact, the one that times out and gets us CIRTed (e.g. recently: Sarah McIntosh especially). What sort of time does this add to very large bulk privilege gets (100, 1000, 10000)?

DMcKinnon-mdsol:
@fpcyan sorry, I should have removed that bit from the description. Connor helped me remove that n+1 query a few days ago. It's performant now. Between [bulk_get_privileges](https://github.com/mdsol/dalton/blob/develop/app/models/privilege.rb#L111-L148) and [find_all_by_operator_and_operable](https://github.com/mdsol/dalton/blob/develop/app/models/privilege.rb#L151-L171), I was under the impression that the latter one was the call giving us all the CIRTs. Is that the case?

lib/policy_machine.rb
fpcyan: Was this ever revisited? Or do you intend to revisit later?

lib/policy_machine_storage_adapters/active_record.rb
```diff
@@ -290,11 +278,14 @@ class Object < ObjectAttribute
     end
 
     class Operation < PolicyElement
-      has_and_belongs_to_many :policy_element_associations, class_name: 'PolicyMachineStorageAdapter::ActiveRecord::PolicyElementAssociation', join_table: 'operations_policy_element_associations'
```
fpcyan:
```
def operation_sets
  Assignment.ascendants_of(self).where(type: PolicyMachineStorageAdapter::ActiveRecord::OperationSet.to_s)
```
think this would be useful? It looks like you want it lower in the file.

DMcKinnon-mdsol: Yea, I tried that, but on a perf test on sandbox, one bulk privilege call for the read operation spiked the db to 50% utilization, so I took the line out.

lib/policy_machine_storage_adapters/active_record.rb
```diff
@@ -696,18 +676,31 @@ def batch_pluck(policy_object, query: {}, fields:, config: {}, &blk)
...
     def accessible_objects(user_or_attribute, operation, options = {})
-      operation = class_for_type('operation').find_by_unique_identifier!(operation.to_s) unless operation.is_a?(class_for_type('operation'))
-      permitting_oas = PolicyElement.where(id: operation.policy_element_associations.where(
-        user_attribute_id: user_or_attribute.descendants | [user_or_attribute],
-      ).select(:object_attribute_id))
+      operation_id = operation.try(:unique_identifier) || operation.to_s
+
+      user_attributes = user_or_attribute.descendants | [user_or_attribute]
+      associations = PolicyElementAssociation.where(user_attribute_id: user_attributes)
+      operation_set_ids = associations.map(&:operation_set_id)
```
fpcyan: also pluck here

lib/policy_machine_storage_adapters/active_record.rb
DMcKinnon-mdsol: Whoops! I must be thinking of a different call. I'll throw this in and see how it does.

fpcyan:
I've got nothing else to add. Superficially this LGTM aside from a few needed plucks.

lib/policy_machine.rb
```diff
@@ -177,7 +178,6 @@ def is_privilege_ignoring_prohibitions?(user_or_attribute, operation, object_or_
   # TODO:  might make privilege a class of its own
   def privileges
     privileges = []
-
```
PoloPro: If you're gonna slay useless lines, you should make this a `reduce`. :3

lib/policy_machine/policy_element.rb
```diff
@@ -234,8 +234,8 @@ def prohibition?
     # Return all associations in which this Operation is included
     # Associations are arrays of PM::Attributes.
     def associations
-      pm_storage_adapter.associations_with(self.stored_pe).map do |assoc|
-        PM::Association.new(assoc[0], assoc[1], assoc[2], assoc[3], pm_storage_adapter)
```
PoloPro: I never knew I needed this gone until now

lib/policy_machine_storage_adapters/active_record.rb
```diff
@@ -244,26 +244,14 @@ def self.bulk_link(pairs_hash)
       end
 
       def self.bulk_associate(associations, upsert_buffer)
-        associations.map! do |user_attribute, set_of_operation_objects, operation_set, object_attribute, policy_machine_uuid|
-          [PolicyElementAssociation.new(user_attribute_id: user_attribute.id,
-                                        object_attribute_id: object_attribute.id,
-                                        operation_set_id: operation_set.id),
-           set_of_operation_objects]
+        associations.map! do |user_attribute, operation_set, object_attribute, policy_machine_uuid|
```
PoloPro: Maybe `_` out the policy_machine_uuid, since it's unused?

lib/policy_machine_storage_adapters/active_record.rb
```diff
@@ -307,44 +298,20 @@ class PolicyElementAssociation < ::ActiveRecord::Base
...
-                                            .delete_all
-          OperationsPolicyElementAssociation.import([:policy_element_association_id, :operation_id],
-                                                     new_operation_objects.map { |op| [self.id, op.id] },
-                                                     validate: false)
-        end
-        self.clear_association_cache
-      end
-
-      def self.add_association(user_attribute, set_of_operation_objects, operation_set, object_attribute, policy_machine_uuid)
+      def self.add_association(user_attribute, operation_set, object_attribute, policy_machine_uuid)
```
PoloPro: `policy_machine_uuid` seems unused?

lib/policy_machine_storage_adapters/active_record.rb
```diff
@@ -663,22 +633,32 @@ def transaction(&block)
     # Returns true if the user has the operation on the object
     def is_privilege?(user_or_attribute, operation, object_or_attribute)
       policy_classes_containing_object = policy_classes_for_object_attribute(object_or_attribute)
+      operation_id = operation.try(:unique_identifier) || operation.to_s
+
       if policy_classes_containing_object.count < 2
-        is_privilege_single_policy_class(user_or_attribute, operation, object_or_attribute)
+        accessible_operations(user_or_attribute, object_or_attribute, operation_id).any?
```
PoloPro: I know it's not pretty, but since this is gonna be our bread-and-butter method, maybe 
`!accessible_operations(user_or_attribute, object_or_attribute, operation_id).empty?`

`!a.empty?` is significantly faster than `a.any?`, according to Googlefu

lib/policy_machine_storage_adapters/active_record.rb
```diff
@@ -663,22 +633,32 @@ def transaction(&block)
...
       else
-        is_privilege_multiple_policy_classes(user_or_attribute, operation, object_or_attribute, policy_classes_containing_object)
+        policy_classes_containing_object.all? do |policy_class|
+          accessible_operations(user_or_attribute, object_or_attribute, operation_id).any?
```
PoloPro: Same above, maybe `!...empty?`

lib/policy_machine_storage_adapters/active_record.rb
```diff
@@ -663,22 +633,32 @@ def transaction(&block)
...
     def scoped_privileges(user_or_attribute, object_or_attribute, options = {})
       policy_classes_containing_object = policy_classes_for_object_attribute(object_or_attribute)
-      if policy_classes_containing_object.count < 2
-        scoped_privileges_single_policy_class(user_or_attribute, object_or_attribute, options)
-      else
-        scoped_privileges_multiple_policy_classes(user_or_attribute, object_or_attribute, policy_classes_containing_object, options)
-      end
+
+      operations =
+        if policy_classes_containing_object.count < 2
```
PoloPro: `.size`?

lib/policy_machine_storage_adapters/active_record.rb
```diff
@@ -663,22 +633,32 @@ def transaction(&block)
...
+          accessible_operations(user_or_attribute, object_or_attribute)
+        else
+          policy_classes_containing_object.flat_map do |policy_class|
+            accessible_operations(user_or_attribute, policy_class.ancestors)
+          end
+        end
+
+      options[:order] ? operations.sort : operations
```
PoloPro: What is default sort order? May be worth defining?

lib/policy_machine_storage_adapters/active_record.rb
```diff
@@ -696,18 +676,31 @@ def batch_pluck(policy_object, query: {}, fields:, config: {}, &blk)
...
+
       if inclusion = options[:includes]
         direct_scope = Adapter.apply_include_condition(scope: direct_scope, key: options[:key], value: inclusion, klass: class_for_type('object'))
         indirect_scope = Adapter.apply_include_condition(scope: indirect_scope, key: options[:key], value: inclusion, klass: class_for_type('object'))
       end
+
       candidates = direct_scope | indirect_scope
-      if options[:ignore_prohibitions] || !(prohibition = class_for_type('operation').find_by_unique_identifier("~#{operation.unique_identifier}"))
+
+      if options[:ignore_prohibitions] || !(prohibition = prohibition_for(operation_id))
```
PoloPro: +1 for cleaning that up

lib/policy_machine_storage_adapters/active_record/postgresql.rb
```diff
@@ -13,34 +13,84 @@ def self.transitive_closure?(ancestor, descendant)
...
+                FROM assignments
+                INNER JOIN assignments_recursive
+                ON assignments_recursive.child_id = assignments.parent_id
+              )
+            )
+
+            SELECT assignments_recursive.child_id
+            FROM assignments_recursive
+          )',
+          [*element_or_scope].map(&:id))
```
PoloPro: For such a large query, may be _slightly_ more clear to knock the final parenthesis to the next line?

lib/policy_machine_storage_adapters/active_record/postgresql.rb
```diff
@@ -13,34 +13,84 @@ def self.transitive_closure?(ancestor, descendant)
...
-        else
-          PolicyElement.where('"policy_elements"."id" IN (SELECT assignments__recursive.child_id FROM (WITH RECURSIVE "assignments__recursive" AS ( SELECT "assignments"."id", "assignments"."child_id", "assignments"."parent_id" FROM "assignments" WHERE "assignments"."parent_id" in (?) UNION ALL SELECT "assignments"."id", "assignments"."child_id", "assignments"."parent_id" FROM "assignments" INNER JOIN "assignments__recursive" ON "assignments__recursive"."child_id" = "assignments"."parent_id" ) SELECT "assignments__recursive".* FROM "assignments__recursive") AS "assignments__recursive")', element_or_scope.map(&:id))
-        end
+        PolicyElement.where('
```
PoloPro: I think this quotation can be safely knocked to the next line

lib/policy_machine_storage_adapters/active_record.rb
```diff
@@ -631,10 +598,13 @@ def associate_later(user_attribute, set_of_operation_objects, operation_set, obj
     # If no associations are found then the empty array should be returned.
     #
     def associations_with(operation)
-      assocs = operation.policy_element_associations(true).includes(:user_attribute, :operations, :operation_set, :object_attribute).all
+      params = { type: PolicyMachineStorageAdapter::ActiveRecord::OperationSet.to_s }
+      operation_sets = Assignment.ancestors_of(operation).where(params)
+      assocs = PolicyElementAssociation.where(operation_set_id: operation_sets.map(&:id))
```
DMcKinnon-mdsol: Great idea! Changed.

lib/policy_machine.rb
DMcKinnon-mdsol: Sigh. Reduces are taking over. Changed.

lib/policy_machine_storage_adapters/active_record.rb
DMcKinnon-mdsol: Interesting. @CGDS or @HonoreDB was the idea to eventually add a `policy_machine_uuid` column to the policy elements associations table?

lib/policy_machine_storage_adapters/active_record.rb
fpcyan: Really? I was going to suggest `.present?` instead of `.any?` and google told me `.any?` was way faster. so !empty > .any? > .present?/.blank?

lib/policy_machine_storage_adapters/active_record/postgresql.rb
fpcyan: alternatively: heredocs

lib/policy_machine_storage_adapters/active_record.rb
DMcKinnon-mdsol: I'm not sure what the default order is, but refactoring has it limits 😝 . That line was around before this PR, and for now, I'm not sure changing it here is appropriate. We may eventually want to sort by the unique identifier, but I'm not sure. For a later day, unless you feel passionately about it.

lib/policy_machine_storage_adapters/active_record.rb
DMcKinnon-mdsol: To remove a db call since we're not yet using AR chaining. But no worries –  I changed it back.

lib/policy_machine.rb
DMcKinnon-mdsol: Later. It _looks_ to me like this is really only intended to be used in tests. We could use the `filter_operation_set_list_by_assigned_operation` method to remove the n+1 query when the storage adapter is psql, but that would require writing adapter-specific logic into this method. Again, that doesn't seem worth it to me right now (since we and no one in the foreseeable future) will ever use this, so I'm inclined to leave it as is unless objections.

lib/policy_machine_storage_adapters/active_record.rb
DMcKinnon-mdsol: Nope, I was thinking of this call. If we try to get this list of operation sets for the `read` operation, for example, then this could load hundreds of thousands of operation sets. The raw SQL (filter_operation_set_list_by_assigned_operation) in the psql adapter is our way around using this call.

DMcKinnon-mdsol:
@mdsol/team04 any other comments?

fpcyan:
I'm gonna merge this before scrum if there are no comments.

fpcyan:
actually I'll wait for @CGDS's upcoming once over.

lib/policy_machine_storage_adapters/active_record.rb
CGDS: i think long run we were going to split out association/assignment tables entirely per pm

CHANGELOG.md
CGDS: fair enough....

lib/policy_machine.rb
```diff
@@ -127,7 +126,9 @@ def is_privilege_ignoring_prohibitions?(user_or_attribute, operation, object_or_
       raise(ArgumentError, "options[:associations] cannot be empty") if associations.empty?
       raise(ArgumentError, "expected each element of options[:associations] to be a PM::Association") unless associations.all?{|a| a.is_a?(PM::Association)}
 
-      associations.keep_if{ |assoc| assoc.includes_operation?(operation) }
+      associations.keep_if do |association|
+        association.operation_set.connected?(operation)
+      end
       return false if associations.empty?
     else
       associations = operation.associations
```
CGDS: note to self: so we filter a collection of associations, check if its empty, and then reassign...wut.

lib/policy_machine.rb
```diff
@@ -167,28 +168,23 @@ def is_privilege_ignoring_prohibitions?(user_or_attribute, operation, object_or_
...
   def privileges
-    privileges = []
-
-    users.each do |user|
+    users.reduce([]) do |memo, user|
       operations.reject(&:prohibition?).each do |operation|
         objects.each do |object|
           if is_privilege?(user, operation, object)
-            privileges << [user, operation, object]
+            memo << [user, operation, object]
```
CGDS: using `+` would give you ideological purity, but i dont particularly care

lib/policy_machine/association.rb
CGDS: not in the diff, not real, and if you press the memory I'll just edit the commit history :P

lib/policy_machine_storage_adapters/active_record/postgresql.rb
```diff
@@ -13,34 +12,86 @@ def self.transitive_closure?(ancestor, descendant)
...
       def self.descendants_of(element_or_scope)
-        #FIXME: Preloading with to_a seems to be necessary because putting complex sql in start_with can
-        # lead to degenerate performance (noticed in ancestors_of call in accessible_objects)
-        # Ideally, fix the SQL so it's both a single call and performant
-        element_or_scope = [*element_or_scope]
-        case element_or_scope.size
-        when 0
-          PolicyElement.none
-        when 1
-          PolicyElement.where('"policy_elements"."id" IN (SELECT assignments__recursive.child_id FROM (WITH RECURSIVE "assignments__recursive" AS ( SELECT "assignments"."id", "assignments"."child_id", "assignments"."parent_id" FROM "assignments" WHERE "assignments"."parent_id" = ? UNION ALL SELECT "assignments"."id", "assignments"."child_id", "assignments"."parent_id" FROM "assignments" INNER JOIN "assignments__recursive" ON "assignments__recursive"."child_id" = "assignments"."parent_id" ) SELECT "assignments__recursive".* FROM "assignments__recursive") AS "assignments__recursive")', element_or_scope.first.id)
```
CGDS: for singular ids, is the `in` as fast as the `=`?

lib/policy_machine_storage_adapters/active_record/postgresql.rb
```diff
@@ -13,34 +12,86 @@ def self.transitive_closure?(ancestor, descendant)
...
+            SELECT assignments_recursive.child_id
+            FROM assignments_recursive
+          )',
+          [*element_or_scope].map(&:id)
+        )
       end
 
       def self.ancestors_of(element_or_scope)
-        #FIXME: Also, removing the superfluous join of Assignment onto the recursive call is hugely beneficial to performance, but not supported
+        # FIXME: Removing the superfluous join of Assignment onto the recursive call is hugely beneficial to performance, but not supported
```
CGDS: which superfluous join now?

lib/policy_machine_storage_adapters/active_record/postgresql.rb
```diff
@@ -13,34 +12,86 @@ def self.transitive_closure?(ancestor, descendant)
...
+          )',
+          [*element_or_scope].map(&:id)
+        )
       end
 
+      # Returns the operation set IDs from the given list where the operation is
+      # a descendant of the operation set.
+      # TODO: Generalize this so that we can arbitrarily filter recursive assignments calls.
+      def self.filter_operation_set_list_by_assigned_operation(operation_set_ids, operation_id)
+        query = <<-SQL
```
CGDS: style seems inconsistent, above using long strings in single quotes.  Still the mild preference for heredocs, but consistent seems gooder.

lib/policy_machine_storage_adapters/active_record/postgresql.rb
fpcyan: According to this person on the internet, `IN` is just as fast: https://stackoverflow.com/a/44762671

lib/policy_machine_storage_adapters/active_record/postgresql.rb
```diff
@@ -13,34 +12,86 @@ def self.transitive_closure?(ancestor, descendant)
...
+          WITH RECURSIVE assignments_recursive AS (
+            (
+              SELECT parent_id, child_id, ARRAY[parent_id] AS parents
+              FROM assignments
+              WHERE parent_id IN (#{operation_set_ids.join(',')})
```
CGDS: would prefer not to do string interpolation directly in sql query, `exec_query` supports binds iirc

lib/policy_machine_storage_adapters/active_record/postgresql.rb
```diff
@@ -13,34 +12,86 @@ def self.transitive_closure?(ancestor, descendant)
...
+          )
+
+          SELECT parents[1]
+          FROM assignments_recursive
+          JOIN policy_elements
+          ON policy_elements.id = assignments_recursive.child_id
+          WHERE policy_elements.unique_identifier = '#{operation_id}'
+          AND type = 'PolicyMachineStorageAdapter::ActiveRecord::Operation'
+        SQL
+        PolicyElement.connection.exec_query(query).rows.flatten.map(&:to_i)
```
CGDS: why `exec_query` instead of `execute` btw? also, is the result set here going to be big enough that we care about optimizing the `flatten.map` part?

lib/policy_machine_storage_adapters/neography.rb
```diff
@@ -163,7 +163,7 @@ def element_in_machine?(pe)
     ##
     # Add the given association to the policy map.  If an association between user_attribute
     # and object_attribute already exists, then replace it with that given in the arguments.
-    def add_association(user_attribute, set_of_operation_objects, object_attribute, policy_machine_uuid)
+    def add_association(user_attribute, object_attribute, policy_machine_uuid = nil)
```
CGDS: why is this nilable?

lib/policy_machine_storage_adapters/active_record/postgresql.rb
CGDS: kewl

lib/policy_machine.rb
DMcKinnon-mdsol: We don't ever use this code path.

lib/policy_machine_storage_adapters/neography.rb
DMcKinnon-mdsol: We don't use pm uuid in associations. I got rid of it but decided that change was too big for this PR, so I reverted. Missed a reversion.

lib/policy_machine_storage_adapters/active_record/postgresql.rb
CGDS: this got donez

PoloPro:
One big thing and then I'll put a merge timer: in the `.gemspec`, we hardcode the version rather than pulling it from `version.rb`.

We should either reference version.rb or hardcode it to 1.6.0 now.

PoloPro:
Will merge by EOD unless we have other comments. @drozenberg I think you're the last man standing. Any concerns?

DMcKinnon-mdsol:
Thanks for picking this up @CGDS

CGDS:
@DMcKinnon-mdsol we gonna use the version file in #122 after this gets merged

DMcKinnon-mdsol:
Yup, sounds good!

PR #122, opened 2017-08-24
[MCC-???] use existing version file instead of string in gemspec

CGDS:
pretty much says it all on the tin.  Switched the unloaded module declaration to a class declaration since it conflicts with the class load later


PoloPro:
It... failed?

CGDS:
oh, this is the travis infrastructure change that don sidestepped by updating the ruby version.  WIll merge develop in when #121 is in, should resolve

PR #124, opened 2017-08-31
[MCC-321135] Fix bug in accessible objects

DMcKinnon-mdsol:
This PR fixes a bug in the active record storage adapter's `accessible_objects` method. Previously this method was not returning the correct operation set ids because our sql was a little off.

Thanks @CGDS for the fix.

@mdsol/team04 please review.

### DIFF
--- a/lib/policy_machine_storage_adapters/active_record/postgresql.rb
+++ b/lib/policy_machine_storage_adapters/active_record/postgresql.rb
@@ -76,7 +76,7 @@ module PolicyMachineStorageAdapter
             )
             UNION ALL
             (
-              SELECT assignments.parent_id, assignments.child_id, parents || assignments.parent_id
+              SELECT assignments.parent_id, assignments.child_id, (parents || assignments.parent_id)
               FROM assignments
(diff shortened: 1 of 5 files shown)


lib/policy_machine_storage_adapters/active_record/postgresql.rb
```diff
@@ -76,7 +76,7 @@ def self.filter_operation_set_list_by_assigned_operation(operation_set_ids, oper
             )
             UNION ALL
             (
-              SELECT assignments.parent_id, assignments.child_id, assignments.parent_id || parents
+              SELECT assignments.parent_id, assignments.child_id, parents || assignments.parent_id
```
PoloPro: Mind adding some parentheses for clarity?

PoloPro:
Small comment about parentheses for clarity, looks good otherwise.

DMcKinnon-mdsol:
@mdsol/team04 any other comments?

PoloPro:
Will merge in 15min unless other comments. @cgds @drozenberg

PR #126, opened 2017-09-28
Reorganize graph connection AR specs for clarity

PoloPro:
This PR reshuffles and more clearly defines the relationships between test elements, because the previous block of code was nigh unreadable. I tried to be more rigid with object definitions and attributes... e.g. all users are now 'blue', user attributes are 'green', etc.

Should make it easier to introduce graph traversals into the PM down the road. (read: soon)

### DIFF
--- a/spec/policy_machine_storage_adapters/active_record_spec.rb
+++ b/spec/policy_machine_storage_adapters/active_record_spec.rb
@@ -565,7 +565,6 @@ describe 'ActiveRecord' do
 
       context 'a filter is applied' do
         it 'applies a single filter if one is supplied' do
-          # green_children = [user_attributes]
           expect(user_1.children(color: 'green')).to match_array(user_attributes.map(&:stored_pe))
         end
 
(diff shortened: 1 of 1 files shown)


spec/policy_machine_storage_adapters/active_record_spec.rb
```diff
@@ -362,291 +362,282 @@
...
-        before do
-          @ua1.update(color: 'green')
-          @new_ua = @pm.create_user_attribute('new_ua')
-          @new_ua.update(color: 'green')
-          @pm.add_assignment(@u1, @new_ua)
-        end
-
         it 'applies a single filter if one is supplied' do
-          expect(@u1.children(color: 'green')).to contain_exactly(@ua1.stored_pe, @new_ua.stored_pe)
+          # green_children = [user_attributes]
```
DMcKinnon-mdsol: Can we make this a more typical comment so that it doesn't read like code?

DMcKinnon-mdsol:
LGTM. Thanks for doing this, @PoloPro.

DMcKinnon-mdsol:
Merging by EOD unless objections @mdsol/team-04.

PR #127, opened 2017-10-16
[MCC-317560] Graph pluck via method chaining

PoloPro:
This PR contains the trivial implementation of plucking attributes from PolicyElement relatives. It literally chains `pluck` onto the end of the other methods, then zips the result with the attribute names, leading to a result of the form...

```
[{ name: 'bob', profession: 'spartan' }, { name: 'wirt', profession: 'frog' }
```

Please review.

### DIFF
--- a/lib/policy_machine_storage_adapters/active_record.rb
+++ b/lib/policy_machine_storage_adapters/active_record.rb
@@ -178,6 +178,30 @@ module PolicyMachineStorageAdapter
         unfiltered_link_children.where(filters)
       end
 
+      def pluck_from_descendants(filters: {}, fields:)
+        assert_valid_filters!(filters)
+
+        plucked_values = Assignment.descendants_of(self).where(filters).pluck(*fields.map(&:to_sym))
+
+        if fields.size > 1
+          plucked_values.map { |val| Hash[fields.zip(val)] }
+        else
+          plucked_values.map { |val| { fields.first.to_sym => val } }
+        end
+      end
+
+      def pluck_from_ancestors(filters: {}, fields:)
+        assert_valid_filters!(filters)
+
+        plucked_values = Assignment.ancestors_of(self).where(filters).pluck(*fields.map(&:to_sym))
+
+        if fields.size > 1
+          plucked_values.map { |val| Hash[fields.zip(val)] }
+        else
+          plucked_values.map { |val| { fields.first.to_sym => val } }
+        end
+      end
+
       def self.serialize(store:, name:, serializer: nil)
         active_record_serialize store, serializer
 
--- a/lib/policy_machine_storage_adapters/active_record.rb
+++ b/lib/policy_machine_storage_adapters/active_record.rb
@@ -191,6 +191,8 @@ module PolicyMachineStorageAdapter
         link_children
       ).each do |graph_method|
         define_method("pluck_from_#{graph_method}") do |filters: {}, fields:|
+          raise(ArgumentError.new("Must provide at least one field to pluck")) unless fields.present?
+
           assert_valid_attributes!(filters.keys)
           assert_valid_attributes!(fields)
 
--- a/lib/policy_machine_storage_adapters/active_record.rb
+++ b/lib/policy_machine_storage_adapters/active_record.rb
@@ -139,42 +139,42 @@ module PolicyMachineStorageAdapter
       end
 
       def descendants(filters = {})
-        assert_valid_filters!(filters)
+        assert_valid_attributes!(filters.keys)
         Assignment.descendants_of(self).where(filters)
       end
 
       def ancestors(filters = {})
-        assert_valid_filters!(filters)
+        assert_valid_attributes!(filters.keys)
         Assignment.ancestors_of(self).where(filters)
       end
 
       def parents(filters = {})
-        assert_valid_filters!(filters)
+        assert_valid_attributes!(filters.keys)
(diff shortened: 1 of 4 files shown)


lib/policy_machine_storage_adapters/active_record.rb
```diff
@@ -139,45 +139,72 @@ def store_attributes
...
+        parents
+        children
+        link_descendants
+        link_ancestors
+        link_parents
+        link_children
+      ).each do |graph_method|
+        define_method("pluck_from_#{graph_method}") do |filters: {}, fields:|
+          assert_valid_attributes!(filters.keys)
+          assert_valid_attributes!(fields)
```
DMcKinnon-mdsol: This and the rest of these methods may return a pretty cryptic error if `fields` is anything other than a list of objects that respond to `to_sym`. Do we check anywhere that `fields` is appropriately populated?

lib/policy_machine_storage_adapters/active_record.rb
```diff
@@ -139,45 +139,72 @@ def store_attributes
...
+
+          plucked_values = public_send(graph_method, filters).pluck(*fields)
+
+          if fields.size > 1
```
DMcKinnon-mdsol: Do we need this conditional? Doesn't
```ruby
plucked_values.map { |values| HashWithIndifferentAccess[fields.zip(values)] }
```
satisfy both cases?

spec/policy_machine_storage_adapters/active_record_spec.rb
```diff
@@ -451,6 +451,45 @@
...
+    describe '#pluck_from_descendants' do
+      context 'no filter is applied' do
+        it 'returns appropriate descendants and the specified attribute' do
+          plucked_results = [{ color: 'green' }, { color: 'green' }, { color: 'green' }]
+          expect(user_1.pluck_from_descendants(fields: [:color])).to match_array(plucked_results)
+        end
+
+        it 'returns appropriate descendants and multiple specified attributes' do
+          plucked_results = [
+            { unique_identifier: 'user_attr_1', color: 'green' },
```
DMcKinnon-mdsol: Mind making at least one of these not `green` so that we can see it's actually associating the correct attributes and values?

DMcKinnon-mdsol: Same comment below.

spec/policy_machine_storage_adapters/active_record_spec.rb
```diff
@@ -504,6 +543,45 @@
...
+      context 'a filter is applied' do
+        it 'applies a single filter if one is supplied' do
+          plucked_results = [{ color: 'blue' }, { color: 'blue' }, { color: 'blue' }]
+          expect(user_attr_1.pluck_from_ancestors(fields: [:color], filters: { color: 'blue' }))
+            .to match_array(plucked_results)
+        end
+
+        it 'applies multiple filters if they are supplied' do
+          args = { fields: [:unique_identifier], filters: { unique_identifier: 'user_1', color: 'blue' } }
+          expect(user_attr_1.pluck_from_ancestors(args)).to contain_exactly({ unique_identifier: 'user_1' })
```
DMcKinnon-mdsol: Do you need the brackets here?

DMcKinnon-mdsol:
Thanks Wes. This is looking pretty good. Mind also adding specs for each of these for when no fields are supplied and no fields or filters are supplied for both when the key is absent and the array is empty?
Also, for a when one of the filtered values is not a string.

lib/policy_machine_storage_adapters/active_record.rb
PoloPro: No, it fails in the singular case

lib/policy_machine_storage_adapters/active_record.rb
PoloPro: I'm not sure I follow? Most of these PM methods will fail if you pass in random arguments. `assert_valid_attributes!` will just tell you if your fields or filters are actually db columns.

PoloPro:
I'm not sure all of those permutations have value as specs. I added a spec ensuring that `fields` has values, as the behavior in the `nil` case is weird.

DMcKinnon-mdsol:
@mdsol/team04 any other comments here? If not, I'll merge by EOD.

fpcyan:
the AR implementation of `pluck` returns an ordered array of values, not an array of hashes. Could this not be confusing down the line?

fpcyan:
That's not blocking btw. I'm days late on commenting about the generalities of the implementation.

PoloPro:
`.select` instantiates an ActiveRecord object for each returned row, and returns a Relation; `.pluck` just returns the attributes in an array.

- `.select` has too much object overhead, I would only use it if I was performing further filtering on the returned attributes.
- `.pluck` returning an array is sort of annoying. At some point in the method chain, someone is going to have to match up the passed in `fields` with the returned array and find out which value corresponds to which field. I feel that's best done at the lowest level.

DMcKinnon-mdsol:
@PoloPro over here like "My api is better than AR's y'all."

@fpcyan any other comments?

DMcKinnon-mdsol:
@fpcyan ping

DMcKinnon-mdsol:
In retrospect, this should have been a minor version update instead of a patch. Let's make sure to update the minor version for future api changes.

PR #128, opened 2017-10-24
Pluck attributes from ancestors of ancestors

PoloPro:
First pass at implementing a 'pluck attributes from ancestors of ancestors' method. Final goal is an single call to load all building block attributes of a ConfigurationType in Dalton, and pass those attributes to the proper CTR parents for memoization.

Edit, 10/26: got it all reduced to a single SQL query, with minimal in-memory parsing. The return value is very SQLish, but that parsing can be handled on the Dalton side (where the proper types of each column are known, e.g. tags is an array, extra_attributes is JSON, etc).

### DIFF
--- a/lib/policy_machine_storage_adapters/active_record.rb
+++ b/lib/policy_machine_storage_adapters/active_record.rb
@@ -207,11 +207,13 @@ module PolicyMachineStorageAdapter
         end
       end
 
-      # This method plucks the attributes of your ancestors' ancestors. If an ancestor does not have any
-      # ancestors of its own, this method will not return it.
+      # This method plucks the attributes of your ancestors' ancestors.
       def pluck_ancestor_attributes_from_ancestors(filters: {}, fields:)
         assert_valid_attributes!(filters.keys)
         assert_valid_attributes!(fields)
+
+        result = Assignment.select_ancestor_ids([id])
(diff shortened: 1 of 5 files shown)


DMcKinnon-mdsol:
Before we start tweaking this PR, do we have a high-level idea of the performance impact here yet? If it works, then is it worth deploying it to sandbox to do some api performance tests?

DMcKinnon-mdsol:
We'll need to update the version & changelog again.

PoloPro:
I deployed a branch to Sandbox that uses this method. The performance is pretty great; looks like CTR#index response time is capped at 1s.

DMcKinnon-mdsol:
Can you also open the dalton-side PR with a DNM merge tag just so we can see the entire change?

lib/policy_machine_storage_adapters/active_record.rb
```diff
@@ -207,6 +207,54 @@ def link_children(filters = {})
         end
       end
 
+      # This method plucks the attributes of your ancestors' ancestors. If an ancestor does not have any
+      # ancestors of its own, this method will not return it.
```
DMcKinnon-mdsol: "will not return it" should be "will not return its attributes", yes?

PR #131, opened 2017-11-09
[MCC-343261] Pluck ancestor tree with attributes

PoloPro:
The goal is to, from a root node, pluck all the attributes from ancestors of ancestors. This is to bypass the memory overhead of instantiating ancestor objects to determine inherited attributes, and remove the n+1 problem inherent in `pluck`ing from each ancestor in turn.

This will be PM version 1.7.0.

### DIFF
--- a/lib/policy_machine_storage_adapters/active_record.rb
+++ b/lib/policy_machine_storage_adapters/active_record.rb
@@ -207,49 +207,28 @@ module PolicyMachineStorageAdapter
         end
       end
 
-      # This method plucks the attributes of your ancestors' ancestors. If an ancestor does not have any
-      # ancestors of its own, this method will not return it.
-      def pluck_ancestor_attributes_from_ancestors(filters: {}, fields:)
+      # This method plucks the attributes of your ancestors' ancestors, along with the
+      # relationships among those ancestors.
+      def pluck_ancestor_tree(filters: {}, fields:)
+        raise(ArgumentError.new("Must provide at least one field to pluck")) unless fields.present?
+
         assert_valid_attributes!(filters.keys)
         assert_valid_attributes!(fields)
 
-        # (1) Database call number 1: obtain the edges of the ancestor 'subtree', represented as a
-        #     hash where each ancestor's id is the key, and the value is an array of _its_ ancestors
-        id_tree = Assignment.select_ancestor_ids([id]).each_with_object({}) do |row, memo|
-          ancestor_array = row['ancestor_ids'].tr('{}','').split(',')
-          ancestor_array.each { |ancestor_id| memo[ancestor_id] ||= [] }
-          memo[row['id']] = ancestor_array
-        end
-
-        # (2) Add :id and :unique_identifier fields to the pluck; and remove the root node's id
-        #     from the ids-to-pluck, as it may not satisfy the filter
+        id_tree = get_ancestor_id_tree([id])
         id_tree.delete(id.to_s)
-        fields_to_pluck = [:id, :unique_identifier] | fields
 
-        # (3) Database call number 2: pluck the specified fields from the root node's ancestors
+        fields_to_pluck = [:id, :unique_identifier] | fields
         plucked_policy_elements = PolicyElement.where(id: id_tree.keys).where(filters).pluck(*fields_to_pluck)
 
-        # (4) Convert the plucked attributes into an attribute hash and merge it into the id subtree
-        attribute_fields = fields_to_pluck - [:id]
-        plucked_policy_elements.each do |policy_element|
-          pe_id = policy_element[0].to_s
-          # Convert [1, "blue", "user_1"] into { color: "blue", uuid: "user_1" }
-          attribute_hash = HashWithIndifferentAccess[attribute_fields.zip(policy_element.drop(1))]
-          id_tree[pe_id] = attribute_hash.merge(ancestor_ids: id_tree[pe_id])
-        end
+        # Convert the plucked attribute arrays into attribute hashes and merge them into the id subtree
+        zip_attributes_into_id_tree!(id_tree, fields_to_pluck - [:id], plucked_policy_elements)
 
-        # (5) For each ancestor, convert its top-level id key to its unique_identifier
-        #     and each of its ancestor ids to the appropriate ancestor attributes
-        id_tree.each_with_object({}) do |(policy_element_id, policy_element_attrs), memo|
+        # For each ancestor hash, convert all instances of 'id' to 'unique_identifier'
+        # and replace each relative's id with that relative's attributes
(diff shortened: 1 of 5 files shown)


lib/policy_machine_storage_adapters/active_record.rb
```diff
@@ -207,6 +207,33 @@ def link_children(filters = {})
         end
       end
 
+      # This method plucks the attributes of your ancestors' ancestors, along with the
```
DMcKinnon-mdsol: That's a curious use of "your" as opposed to "a policy machine's". In the interest of not plucking my ancestors, mild preference for changing.

lib/policy_machine_storage_adapters/active_record/postgresql.rb
```diff
@@ -63,6 +63,33 @@ def self.ancestors_of(element_or_scope)
         PolicyElement.where(query, [*element_or_scope].map(&:id))
       end
 
+      # Return an ActiveRecord::Relation containing the ids of all ancestors and the
+      # interstitial relationships, as a string of ancestor_ids
+      def self.select_ancestor_ids(root_element_ids)
```
DMcKinnon-mdsol: Does it make more sense to call this `find_ancestor_ids`?

lib/policy_machine_storage_adapters/active_record.rb
```diff
@@ -290,6 +317,38 @@ def assert_valid_attributes!(attributes)
           raise ArgumentError, "Provided argument contains invalid keys, valid keys are #{PolicyElement.column_names}"
         end
       end
+
+      # Returns a hash containing the ancestors of the root nodes, with the interstitial
+      # ancestor relationships represented by the key/value pairs.
+      def get_ancestor_id_tree(root_nodes)
+        Assignment.select_ancestor_ids(root_nodes).each_with_object({}) do |row, memo|
+          id_array = row['ancestor_ids'].tr('{}','').split(',')
+          id_array.each { |ancestor_id| memo[ancestor_id] ||= [] }
```
fpcyan: What's the impetus for this instantiation? It seems like we're going to overwrite it anyway. Is there a situation that would arise where a key in the `ancestor_id_tree` has no relationships?

lib/policy_machine_storage_adapters/active_record.rb
```diff
@@ -290,6 +317,38 @@ def assert_valid_attributes!(attributes)
           raise ArgumentError, "Provided argument contains invalid keys, valid keys are #{PolicyElement.column_names}"
         end
       end
+
+      # Returns a hash containing the ancestors of the root nodes, with the interstitial
+      # ancestor relationships represented by the key/value pairs.
+      def get_ancestor_id_tree(root_nodes)
+        Assignment.select_ancestor_ids(root_nodes).each_with_object({}) do |row, memo|
```
fpcyan: I'd appreciate it if for `each_with_objects` that return the result to have a more descriptive name than `memo`.

i.e. This `memo` is the `ancestor_id_tree`. In `pluck_ancestor_tree`, I'm not sure what you'd call it, but a name might help.

spec/policy_machine_storage_adapters/active_record_spec.rb
```diff
@@ -603,6 +603,109 @@
...
+
+        pm1.add_assignment(user_attr_7, user_attr_4)
+        pm1.add_assignment(user_attr_8, user_attr_5)
+        pm1.add_assignment(user_attr_9, user_attr_6)
+      end
+
+      context 'no filter is applied' do
+        it 'returns appropriate ancestors and the specified attribute' do
+          plucked_results = {
+            'user_1' => [],
```
fpcyan: any reason not to use colons instead of hash rockets?

fpcyan: In the method you specifically delete the key for the originator's id. Why is it in its own ancestor tree?

spec/policy_machine_storage_adapters/active_record_spec.rb
```diff
@@ -603,6 +603,109 @@
...
+
+        it 'returns appropriate results when filters apply to ancestors but not their ancestors' do
+          plucked_results = { 'user_attr_4' => [{}], 'user_attr_5' => [{}], 'user_attr_6' => [{}] }
+          params = { fields: [:unique_identifier], filters: { color: 'gold'} }
+          expect(user_attr_1.pluck_ancestor_tree(params)).to match_array(plucked_results)
+        end
+
+        it 'returns appropriate results when filters apply to no ancestors' do
+          params = { fields: [:unique_identifier], filters: { color: 'obsidian'} }
+          expect(user_attr_1.pluck_ancestor_tree(params)).to match_array({})
```
fpcyan: `.to eq([])` is a little clearer.

lib/policy_machine_storage_adapters/active_record.rb
```diff
@@ -290,6 +317,38 @@ def assert_valid_attributes!(attributes)
...
+          id_array = row['ancestor_ids'].tr('{}','').split(',')
+          id_array.each { |ancestor_id| memo[ancestor_id] ||= [] }
+          memo[row['id']] = id_array
+        end
+      end
+
+      # Zip an array of attributes into an id tree, with interstitial relationships preserved
+      # under the "relative_ids" key
+      def zip_attributes_into_id_tree!(id_tree, plucked_fields, plucked_attributes)
```
DMcKinnon-mdsol: Why mutate `id_tree` in place? Can we not do that? I'd prefer to just set the return value to the changed id tree?

lib/policy_machine_storage_adapters/active_record.rb
PoloPro: The situation where the ancestor has no ancestors of its own, i.e. it's a leaf node.

spec/policy_machine_storage_adapters/active_record_spec.rb
PoloPro: The key-value pairs are formed from the `child_id`-`parent_id` relationship. `assignments_recursive` starts by specifying the first `child_id` as the root node. Thus it is returned as a key in its own ancestor tree. I manually delete it, since a) it is implied by calling the method on it, and b) it doesn't necessarily satisfy the filters on its ancestors.

Perfect example: `ct.pluck_ancestor_tree(filters: { active_policy_element_type: 'ConfigurationTypeRole' }, fields: [:name, :tags])` The ConfigurationType doesn't satisfy the CTR ape_type filter, so we nix it.

DMcKinnon-mdsol:
I have quite a few ideas for making this PR more aesthetic, but all have unfortunately also come with a performance hit. I'll keep staring at this, and we can follow up in a separate PR if we come up with anything to improve it.

@fpcyan @drozenberg @CGDS merging in 1hr unless objections.

PR #136, opened 2018-03-13
[MCC-361558] Add accessible_ancestor_objects to ActiveRecord storage adapter

PoloPro:
This PR adds a `accessible_ancestor_objects` method to the ActiveRecord storage adapter. This method adds object scoping to the existing `accessible_objects` method. An object is only "accessible" if it is an ancestor of the specified root object.

### DIFF
--- a/lib/policy_machine.rb
+++ b/lib/policy_machine.rb
@@ -271,9 +271,8 @@ class PolicyMachine
     if policy_machine_storage_adapter.respond_to?(:accessible_ancestor_objects)
       policy_machine_storage_adapter.accessible_ancestor_objects(user_or_attribute, operation, root_object, options)
     else
-      Warn.once "WARNING: accessible_ancestor_objects is not implemented for storage adapter " \
-                "#{policy_machine_storage_adapter.class}. Returning all accessible_objects."
-      accessible_objects(user_or_attribute, operation, options)
+      raise NoMethodError, "accessible_ancestor_objects is not implemented for storage adapter " \
+                           "#{policy_machine_storage_adapter.class}."
     end
   end
 
--- a/lib/policy_machine.rb
+++ b/lib/policy_machine.rb
@@ -267,6 +267,16 @@ class PolicyMachine
     end
   end
 
+  def accessible_ancestor_objects(user_or_attribute, operation, root_object, options = {})
+    if policy_machine_storage_adapter.respond_to?(:accessible_ancestor_objects)
+      policy_machine_storage_adapter.accessible_ancestor_objects(user_or_attribute, operation, root_object, options)
+    else
+      Warn.once "WARNING: accessible_ancestor_objects is not implemented for storage adapter " \
+                "#{policy_machine_storage_adapter.class}. Returning all accessible_objects."
+      accessible_objects(user_or_attribute, operation, options)
+    end
+  end
+
   ##
   # Returns an array of all user_attributes a PM::User is assigned to,
   # directly or indirectly.
--- a/lib/policy_machine_storage_adapters/active_record.rb
+++ b/lib/policy_machine_storage_adapters/active_record.rb
@@ -797,21 +797,29 @@ module PolicyMachineStorageAdapter
     end
 
     # Version of accessible_objects which only returns objects that are
-    # descendants of a specified root object.
-    def accessible_descendant_objects(user_or_attribute, operation, root_object, options = {})
-      operation_id = operation.try(:unique_identifier) || operation.to_s
+    # ancestors of a specified root object or the object itself.
+    def accessible_ancestor_objects(user_or_attribute, operation, root_object, options = {})
+      # The final set of accessible objects must be contained in the following set
+      ancestor_objects = root_object.ancestors(type: class_for_type('object')) + [root_object.stored_pe]
 
-      user_attributes = user_or_attribute.descendants | [user_or_attribute]
-      associations = PolicyElementAssociation.where(user_attribute_id: user_attributes.map(&:id))
+      # Fetch all of the PEAs using the given UA or its descendants
+      user_attribute_ids = user_or_attribute.descendants.pluck(:id) | [user_or_attribute.id]
+      associations = PolicyElementAssociation.where(user_attribute_id: user_attribute_ids)
       operation_set_ids = associations.pluck(:operation_set_id)
 
+      # Narrow the list of PEAs to just those containing the specified operation
+      operation_id = operation.try(:unique_identifier) || operation.to_s
       filtered_operation_set_ids = Assignment.filter_operation_set_list_by_assigned_operation(operation_set_ids, operation_id)
       filtered_associations =
         associations.select do |association|
           filtered_operation_set_ids.include?(association.operation_set_id)
         end
 
-      permitting_oas = PolicyElement.where(id: filtered_associations.map(&:object_attribute_id))
+      # TODO: Add a filter here to limit the list of permitting OAs to the set of ancestor objects
+      # Complication: an object may be accessible from ancestors not in the set
+      # i.e. how interconnected are subgraphs?
+      permitting_oa_ids = filtered_associations.map(&:object_attribute_id)
+      permitting_oas = PolicyElement.where(id: permitting_oa_ids)
 
       direct_scope = permitting_oas.where(type: class_for_type('object'))
       indirect_scope = Assignment.ancestors_of(permitting_oas).where(type: class_for_type('object'))
@@ -821,7 +829,8 @@ module PolicyMachineStorageAdapter
         indirect_scope = Adapter.apply_include_condition(scope: indirect_scope, key: options[:key], value: inclusion, klass: class_for_type('object'))
       end
 
-      candidates = (direct_scope | indirect_scope) | root_object.descendants
+      # Ensure all candidates are ancestors of the specified root_object
+      candidates = (direct_scope | indirect_scope) & ancestor_objects
 
       if options[:ignore_prohibitions] || !(prohibition = prohibition_for(operation_id))
         candidates
--- a/lib/policy_machine_storage_adapters/active_record.rb
+++ b/lib/policy_machine_storage_adapters/active_record.rb
@@ -807,7 +807,7 @@ module PolicyMachineStorageAdapter
 
       # Short-circuit and return all ancestors (minus prohibitions), if authorized on the root node
       if is_privilege?(user_or_attribute, operation, root_object_pe)
-        if options[:ignore_prohibitions]
+        if options[:ignore_prohibitions] || !(prohibition = prohibition_for(operation))
           if inclusion = options[:includes]
             return ancestor_objects.select { |obj| obj.send(options[:key].to_sym).include?(inclusion) }
           else
@@ -837,9 +837,6 @@ module PolicyMachineStorageAdapter
           filtered_operation_set_ids.include?(association.operation_set_id)
         end
 
-      # TODO: Add a filter here to limit the list of permitting OAs to the set of ancestor objects
-      # Complication: an object may be accessible from ancestors not in the set
-      # i.e. how interconnected are subgraphs?
       permitting_oa_ids = filtered_associations.map(&:object_attribute_id)
       permitting_oas = PolicyElement.where(id: permitting_oa_ids)
 
--- a/lib/policy_machine_storage_adapters/active_record.rb
+++ b/lib/policy_machine_storage_adapters/active_record.rb
@@ -821,7 +821,7 @@ module PolicyMachineStorageAdapter
         indirect_scope = Adapter.apply_include_condition(scope: indirect_scope, key: options[:key], value: inclusion, klass: class_for_type('object'))
       end
 
-      candidates = direct_scope | indirect_scope
+      candidates = (direct_scope | indirect_scope) | root_object.descendants
 
       if options[:ignore_prohibitions] || !(prohibition = prohibition_for(operation_id))
         candidates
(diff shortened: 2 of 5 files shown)


lib/policy_machine_storage_adapters/active_record.rb
```diff
@@ -796,6 +796,49 @@ def accessible_objects(user_or_attribute, operation, options = {})
...
     end
 
+    # Version of accessible_objects which only returns objects that are
+    # ancestors of a specified root object or the object itself.
+    def accessible_ancestor_objects(user_or_attribute, operation, root_object, options = {})
+      # The final set of accessible objects must be contained in the following set
+      ancestor_objects = root_object.ancestors(type: class_for_type('object')) + [root_object.stored_pe]
+
+      # Fetch all of the PEAs using the given UA or its descendants
+      user_attribute_ids = user_or_attribute.descendants.pluck(:id) | [user_or_attribute.id]
```
fpcyan: There's a lot of code duplication between this and `accessible_objects` which could be DRYed up.

lib/policy_machine_storage_adapters/active_record.rb
```diff
@@ -796,6 +796,49 @@ def accessible_objects(user_or_attribute, operation, options = {})
...
+      associations = PolicyElementAssociation.where(user_attribute_id: user_attribute_ids)
+      operation_set_ids = associations.pluck(:operation_set_id)
+
+      # Narrow the list of PEAs to just those containing the specified operation
+      operation_id = operation.try(:unique_identifier) || operation.to_s
+      filtered_operation_set_ids = Assignment.filter_operation_set_list_by_assigned_operation(operation_set_ids, operation_id)
+      filtered_associations =
+        associations.select do |association|
```
fpcyan: `associations.where(operation_set_id: filtered_operation_set_ids)`

lib/policy_machine.rb
```diff
@@ -267,6 +267,16 @@ def accessible_objects(user_or_attribute, operation, options = {})
     end
   end
 
+  def accessible_ancestor_objects(user_or_attribute, operation, root_object, options = {})
+    if policy_machine_storage_adapter.respond_to?(:accessible_ancestor_objects)
+      policy_machine_storage_adapter.accessible_ancestor_objects(user_or_attribute, operation, root_object, options)
+    else
+      Warn.once "WARNING: accessible_ancestor_objects is not implemented for storage adapter " \
+                "#{policy_machine_storage_adapter.class}. Returning all accessible_objects."
+      accessible_objects(user_or_attribute, operation, options)
```
hstebbins-mdsol: Should returning all accessible objects be the default behavior if the method isn't implemented, or should it raise? 

I guess my concern is that someone could use this method and work on the return assuming that it is scoped to _only_ ancestor objects, which could lead to Bad Things if they don't also verify that each returned item is an ancestor.

spec/policy_machine_storage_adapters/active_record_spec.rb
```diff
@@ -407,6 +407,147 @@
...
+        ado_pm.add_association(ua, not_reader, parent_fish)
+        all_accessible_from_grandparent = %w(grandparent_fish uncle_fish cousin_fish)
+
+        expect(ado_pm.accessible_ancestor_objects(u1, read, grandparent_fish, key: :unique_identifier).map(&:unique_identifier))
+          .to match_array(all_accessible_from_grandparent)
+        expect(ado_pm.accessible_ancestor_objects(u1, write, parent_fish, key: :unique_identifier).map(&:unique_identifier))
+          .to contain_exactly('child_fish_1')
+      end
+
+      it 'does not return objects which prohibited by an out-of-scope descendant' do
```
hstebbins-mdsol: typo in this description?

lib/policy_machine_storage_adapters/active_record.rb
```diff
@@ -772,6 +772,65 @@ def accessible_objects(user_or_attribute, operation, options = {})
...
+    # ancestors of a specified root object or the object itself
+    def accessible_ancestor_objects(user_or_attribute, operation, root_object, options = {})
+      # Drop down to the root's stored policy element if not already there
+      root_object_pe = root_object.respond_to?(:stored_pe) ? root_object.stored_pe : root_object
+
+      # The final set of accessible objects must be contained in the following set
+      ancestor_objects = root_object_pe.ancestors(type: class_for_type('object')) + [root_object_pe]
+
+      # Short-circuit and return all ancestors (minus prohibitions), if authorized on the root node
+      if is_privilege?(user_or_attribute, operation, root_object_pe)
```
hstebbins-mdsol: How about extracting parts of this block into a method(s)? It seems like there's a lot of similarity with the functionality in different parts of the logic tree.

lib/policy_machine_storage_adapters/active_record.rb
```diff
@@ -772,6 +772,65 @@ def accessible_objects(user_or_attribute, operation, options = {})
...
+        if options[:ignore_prohibitions] || !(prohibition = prohibition_for(operation))
+          if inclusion = options[:includes]
+            return ancestor_objects.select { |obj| obj.send(options[:key].to_sym).include?(inclusion) }
+          else
+            return ancestor_objects
+          end
+        else
+          ancestor_objects_minus_prohibitions = ancestor_objects - accessible_ancestor_objects(user_or_attribute, prohibition_for(operation), root_object_pe)
```
hstebbins-mdsol: I'm wondering if it's worth it (or works at all) to have an optional ancestor_objects parameter that can be passed to prevent the recursion here from hitting the db again in line#782.

lib/policy_machine_storage_adapters/active_record.rb
```diff
@@ -772,6 +772,65 @@ def accessible_objects(user_or_attribute, operation, options = {})
...
+      root_object_pe = root_object.respond_to?(:stored_pe) ? root_object.stored_pe : root_object
+
+      # The final set of accessible objects must be contained in the following set
+      ancestor_objects = root_object_pe.ancestors(type: class_for_type('object')) + [root_object_pe]
+
+      # Short-circuit and return all ancestors (minus prohibitions), if authorized on the root node
+      if is_privilege?(user_or_attribute, operation, root_object_pe)
+        if options[:ignore_prohibitions] || !(prohibition = prohibition_for(operation))
+          if inclusion = options[:includes]
+            return ancestor_objects.select { |obj| obj.send(options[:key].to_sym).include?(inclusion) }
```
DMcKinnon-mdsol: It doesn't look like we need any of these hard returns.

lib/policy_machine_storage_adapters/active_record.rb
```diff
@@ -772,6 +772,65 @@ def accessible_objects(user_or_attribute, operation, options = {})
       end
     end
 
+    # Version of accessible_objects which only returns objects that are
+    # ancestors of a specified root object or the object itself
+    def accessible_ancestor_objects(user_or_attribute, operation, root_object, options = {})
+      # Drop down to the root's stored policy element if not already there
```
DMcKinnon-mdsol: I don't think the meaning of "drop down" is immediately obvious to someone not already intimately familiar with all of the pm. Can we reword this?

lib/policy_machine_storage_adapters/active_record.rb
```diff
@@ -772,6 +772,65 @@ def accessible_objects(user_or_attribute, operation, options = {})
...
+      root_object_pe = root_object.respond_to?(:stored_pe) ? root_object.stored_pe : root_object
```
DMcKinnon-mdsol: Two things:
- `root_object` is a pe either way, right? Since we want it to be a _stored_ pe here, I think `root_object_stored_pe` is a better name here.

- This works:
```ruby
 root_object_pe = root_object.try(:stored_pe) || root_object
```

DMcKinnon-mdsol: Also, testing what `root_object` responds to seems a roundabout way to get at what we actually want to figure out here, which is if it's of type [whatever stored_pe is]. If you'd prefer to keep the code similar to what's currently there, I'd advocate for `is_a?` over `responds_to?`.

lib/policy_machine.rb
PoloPro: The warning should address that, but there is an argument to be made that it should return `[]` or error out. @mdsol/team04 any opinions?

DMcKinnon-mdsol: Having it default to `accessible_objects` is disingenuous, and I relying on people to read errors is...not reliable. I could go either way on this, but I think we should just let it raise if it's not implemented.

lib/policy_machine_storage_adapters/active_record.rb
```diff
@@ -772,6 +772,63 @@ def accessible_objects(user_or_attribute, operation, options = {})
...
+      # Short-circuit and return all ancestors (minus prohibitions) if the user_or_attribute is authorized on the root node
+      if is_privilege?(user_or_attribute, operation, root_object_stored_pe)
+        apply_include_condition!(ancestor_objects, options[:key], options[:includes])
+
+        if !options[:ignore_prohibitions] && prohibition = prohibition_for(operation)
+          prohibited_ancestor_objects = accessible_ancestor_objects(
+            user_or_attribute,
+            prohibition,
+            root_object_stored_pe,
+            options.merge(ignore_prohibitions: true)
```
DMcKinnon-mdsol: Don't we _not_ want to ignore prohibitions here?

lib/policy_machine_storage_adapters/active_record.rb
```diff
@@ -772,6 +772,63 @@ def accessible_objects(user_or_attribute, operation, options = {})
...
+    def accessible_ancestor_objects(user_or_attribute, operation, root_object, options = {})
+      # If the root_object is a generic PM::Object, convert it the appropriate storage adapter Object
+      root_object_stored_pe = root_object.try(:stored_pe) || root_object
+
+      # The final set of accessible objects must be ancestors of the root_object
+      ancestor_objects = root_object_stored_pe.ancestors(type: class_for_type('object')) + [root_object_stored_pe]
+
+      # Short-circuit and return all ancestors (minus prohibitions) if the user_or_attribute is authorized on the root node
+      if is_privilege?(user_or_attribute, operation, root_object_stored_pe)
+        apply_include_condition!(ancestor_objects, options[:key], options[:includes])
```
DMcKinnon-mdsol: I'm looking for ways to make this `accessible_ancestor_objects` method more palatable. @PoloPro what do you think about extracting the body of this `if is_privilege?(user_or_attribute, operation, root_object_stored_pe)` block into a separate private method?

lib/policy_machine_storage_adapters/active_record.rb
```diff
@@ -772,6 +772,63 @@ def accessible_objects(user_or_attribute, operation, options = {})
...
+      user_attribute_ids = user_or_attribute.descendants.pluck(:id) | [user_or_attribute.id]
+      associations = PolicyElementAssociation.where(user_attribute_id: user_attribute_ids)
+      operation_set_ids = associations.pluck(:operation_set_id)
+
+      # Narrow the list of PEAs to just those containing the specified operation
+      operation_id = operation.try(:unique_identifier) || operation.to_s
+      filtered_operation_set_ids = Assignment.filter_operation_set_list_by_assigned_operation(operation_set_ids, operation_id)
+      filtered_associations = associations.where(operation_set_id: filtered_operation_set_ids)
+
+      permitting_oa_ids = filtered_associations.map(&:object_attribute_id)
```
DMcKinnon-mdsol: Why do a map here but a pluck above? Does pluck construct its own sql query above? If so, we should probably use map in both places.

lib/policy_machine_storage_adapters/active_record.rb
```diff
@@ -772,6 +772,63 @@ def accessible_objects(user_or_attribute, operation, options = {})
...
+      permitting_oas = PolicyElement.where(id: permitting_oa_ids)
+
+      direct_scope = permitting_oas.where(type: class_for_type('object'))
```
DMcKinnon-mdsol: Comment please

lib/policy_machine_storage_adapters/active_record.rb
```diff
@@ -772,6 +772,63 @@ def accessible_objects(user_or_attribute, operation, options = {})
...
+      indirect_scope = Assignment.ancestors_of(permitting_oas).where(type: class_for_type('object'))
+
+      if inclusion = options[:includes]
```
DMcKinnon-mdsol: Comment please

lib/policy_machine_storage_adapters/active_record.rb
PoloPro: This call uses the prohibition, if one exists. We don't want to continue the recursion from "~read" to "~~read".

lib/policy_machine_storage_adapters/active_record.rb
PoloPro: I'll `pluck` in both.

lib/policy_machine_storage_adapters/active_record.rb
```diff
@@ -952,6 +1001,31 @@ def accessible_operations(user_or_attribute, object_or_attribute, operation_id =
...
+    def all_ancestor_objects(user_or_attribute, operation, root_object, ancestor_objects, options)
+      apply_include_condition!(ancestor_objects, options[:key], options[:includes])
+
+      if !options[:ignore_prohibitions] && prohibition = prohibition_for(operation)
+        prohibited_ancestor_objects = accessible_ancestor_objects(
+          user_or_attribute,
+          prohibition,
+          root_object,
+          options.merge(ignore_prohibitions: true)
+        )
```
hstebbins-mdsol: I might be reading this wrong, but could you add an optional parameter to `accessible_ancestor_objects` that can take the already-loaded ancestor_objects here to prevent requesting them again in https://github.com/mdsol/the_policy_machine/pull/136/files#diff-aae96e1c7c46150cd4f06b06c69b75d3R782?

hstebbins-mdsol:
LGTM!

hstebbins-mdsol:
Merging in 1h unless objections @mdsol/team04

PR #137, opened 2018-03-22
Downversioned 'pg' gem to 0.15.0

PoloPro:
Due to an error in the pg gem, v1.0.0 and its interaction with core Rails, we'll need to enforce a lower pg version for the PolicyMachine.

Once the error is patched will look into restoring the gem to v1.0.0+.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,8 @@
 # Changelog
 
+## 1.7.1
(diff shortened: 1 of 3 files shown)


DMcKinnon-mdsol:
Aww so sad. Thanks, Wes. Merging by EOD unless objections @mdsol/team-04.

PR #141, opened 2018-06-18
[MCC-396371] Feature/incompatibility check

drozenberg:
The new conflicting roles check requires a sql query to be run on a role create/update and a grant create/update to determine if any incompatibilities exist.

These sql queries are introduce to the policy machine in this PR. The idea behind these queries basically checks if the update to a role or grant results in a user being assigned two different building block roles from the same conflicting roles set on a given cascaded path.

@DMcKinnon-mdsol @PoloPro @hstebbins-mdsol

### DIFF
--- a/lib/policy_machine_storage_adapters/active_record.rb
+++ b/lib/policy_machine_storage_adapters/active_record.rb
@@ -99,10 +99,12 @@ module PolicyMachineStorageAdapter
     def self.conflicts_for_operables(operable_ids, crs_ids)
       descendants_query = <<-SQL
         WITH RECURSIVE descendants AS (
-          SELECT parent_id, child_id, parent_id as root_node FROM assignments
-          WHERE parent_id IN (#{operable_ids.join(", ")})
+          SELECT parent_id, child_id, parent_id as root_node
+          FROM assignments
+            WHERE parent_id IN (#{operable_ids.join(", ")})
           UNION ALL
-          SELECT a.parent_id, a.child_id, root_node FROM assignments a
+          SELECT a.parent_id, a.child_id, root_node
+          FROM assignments a
           INNER JOIN descendants d 
             ON a.parent_id = d.child_id
             WHERE a.parent_id = d.child_id
@@ -136,13 +138,15 @@ module PolicyMachineStorageAdapter
 
       ancestors_query = <<-SQL
         WITH RECURSIVE ancestors AS (
-          SELECT child_id, parent_id, child_id AS root_node FROM assignments
-          WHERE child_id IN (#{operable_ids.join(", ")})
+          SELECT child_id, parent_id, child_id AS root_node
+          FROM assignments
+            WHERE child_id IN (#{operable_ids.join(", ")})
           UNION ALL
-          SELECT a.child_id, a.parent_id, root_node FROM assignments a
+          SELECT a.child_id, a.parent_id, root_node
+          FROM assignments a
           INNER JOIN ancestors ans
             ON ans.parent_id = a.child_id
-            WHERE ans.parent_id = a.child_id
+          WHERE ans.parent_id = a.child_id
         )
         SELECT ans.root_node, ra.operator_uri, crs.id FROM ancestors ans
         JOIN policy_element_associations pea
--- a/lib/policy_machine_storage_adapters/active_record.rb
+++ b/lib/policy_machine_storage_adapters/active_record.rb
@@ -175,7 +175,7 @@ module PolicyMachineStorageAdapter
     end
 
     # Check conflicts for Grant updates
-    def self.conflicts_for_operators_and_operables(operator_uri, operable_ids)
+    def self.conflicts_for_operators_and_operables(operator_uri, operable_ids, crs_ids)
       descendants_query = <<-SQL
         WITH RECURSIVE descendants AS (
           SELECT parent_id, child_id, parent_id AS root_node FROM assignments
@@ -208,6 +208,7 @@ module PolicyMachineStorageAdapter
           AND os.id = bb_a.parent_id
           AND bb_a.child_id = crs_a.child_id
           AND crs_a.parent_id = crs.id
+          AND crs_a.parent_id IN (#{crs_ids.join(", ")})
         GROUP BY root_node, crs.id
         HAVING COUNT(DISTINCT bb_a.child_id) > 1
       SQL
@@ -242,6 +243,7 @@ module PolicyMachineStorageAdapter
           AND os.id = bb_a.parent_id
           AND bb_a.child_id = crs_a.child_id
           AND crs_a.parent_id = crs.id
+          AND crs_a.parent_id IN (#{crs_ids.join(", ")})
         GROUP BY root_node, crs.id
         HAVING COUNT(DISTINCT bb_a.child_id) > 1
       SQL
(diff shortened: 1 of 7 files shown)


lib/policy_machine_storage_adapters/active_record.rb
```diff
@@ -95,6 +95,161 @@ def self.persist_buffers!
...
     end
 
+    # Check conflicts for Configuration Type Role Updates
+    def self.conflicts_for_operables(operable_ids, crs_ids)
+      descendants_query = <<-SQL
+        WITH RECURSIVE descendants AS (
+          SELECT parent_id, child_id, parent_id as root_node FROM assignments
+          WHERE parent_id IN (#{operable_ids.join(", ")})
+          UNION ALL
+          SELECT a.parent_id, a.child_id, root_node FROM assignments a
```
fpcyan: For the purposes of future maintainability, it may be better to alias things semantically where possible.

lib/policy_machine_storage_adapters/active_record.rb
```diff
@@ -95,6 +95,161 @@ def self.persist_buffers!
...
+        JOIN policy_element_associations AS pea
+          ON pea.object_attribute_id = d.child_id OR pea.object_attribute_id = d.parent_id
+        JOIN policy_elements as ra
+          ON ra.id = pea.user_attribute_id
+        JOIN policy_elements as os
+          ON pea.operation_set_id = os.id
+        JOIN assignments as bb_a
+          ON os.id = bb_a.parent_id
+        JOIN assignments as crs_a
+          ON bb_a.child_id = crs_a.child_id
```
hstebbins-mdsol: There's a potential performance improvement by adding the `IN (#{crs_ids.join(", ")})` clause to the join here as well. 

Unfortunately, Postgres doesn't allow `IN` equality checking in joins (from what I remember, at least). You could build individual equality checks by looping over `crs_ids` and injecting each as an equality clause. Something like this may work: 
```ruby
JOIN policy_elements as crs_a 
ON os.id = bb_a.parent_id 
AND (
#{
  crs_ids.map do |crs_id| 
    "#{crs_id == crs_ids.first ? "" : "OR "}crs_a.parent_id = #{crs_id}"
  end.join(" ")
}
)
```

I'm not sure how big of a performance increase this would have, if any. If possible, we could just test it quickly on some Sandbox data or something.

fpcyan: > Unfortunately, Postgres doesn't allow IN equality checking in joins (from what I remember, at least).

It does, but it's questionable whether it'll be faster or not. psql doesn't seem to like it very much if you do big `IN ()` sets.

DMcKinnon-mdsol:
Do you plan to add any tests for these?

lib/policy_machine_storage_adapters/active_record.rb
```diff
@@ -95,6 +95,161 @@ def self.persist_buffers!
...
+        JOIN policy_elements AS crs
+          ON crs_a.parent_id = crs.id
+        AND crs.active_policy_element_type = 'ConflictingRolesSet'
```
DMcKinnon-mdsol: This is Dalton-specific logic. It doesn't belong in the policy machine.

DMcKinnon-mdsol:
@hstebbins-mdsol I'm fine updating the release tags to be correct in this PR.

lib/policy_machine_storage_adapters/active_record.rb
drozenberg: active_policy_element is indeed Dalton-specific logic. However, pulling this into Dalton would make the conflicting roles check active record adapter-specific logic. There are some downsides to both approaches, so I suggest we move forward as is and have a design meeting when @CGDS gets back about whether or not we should extend the policy machine with a new conflicting roles set policy element.

drozenberg:
@DMcKinnon-mdsol Plenty of tests regarding these queries in Dalton. As we may pull this out of the policy machine, I recommend we keep them there until an alternate decision is made.

DMcKinnon-mdsol:
Hmmm. Okay, between there not being any tests for this here _and_ us not have made a decision around how we're going to handle `active_policy_element_base`, I've changed my mind again (sorry) and don't think we should merge this as is. It's just not quite production ready.

lib/policy_machine_storage_adapters/active_record.rb
hstebbins-mdsol: Yeah, I ran some quick tests and it didn't make a large performance improvement. I don't think it's necessary.

lib/policy_machine_storage_adapters/active_record.rb
```diff
@@ -95,6 +95,179 @@ def self.persist_buffers!
...
+          AND os.id = bb_a.parent_id
+          AND bb_a.child_id = crs_a.child_id
+          AND crs_a.parent_id = crs.id
+          AND crs_a.parent_id IN (#{crs_ids.join(", ")})
+        GROUP BY root_node, ra.operator_uri, crs.id
+        HAVING COUNT(DISTINCT bb_a.child_id) > 1
+      SQL
+
+      result = PolicyElement.connection.exec_query(descendants_query).rows.flatten.map(&:to_i)
+      result.concat(PolicyElement.connection.exec_query(ancestors_query).rows.flatten.map(&:to_i))
```
hstebbins-mdsol: I think we should combine the queries so it only hits the DB once.

hstebbins-mdsol: We could even leave them as separate strings for now without any modifications - we just need to combine the strings. Something like this could work I think: 
```sql
WITH descendants_query AS (
  #{descendants_query}
),
ancestors_query AS (
  #{ancestors_query}
)
SELECT root_node, id 
FROM descendants_query
UNION ALL 
SELECT root_node, id
FROM ancestors_query
```

lib/policy_machine_storage_adapters/active_record.rb
```diff
@@ -95,6 +95,179 @@ def self.persist_buffers!
...
+          FROM assignments
+          WHERE parent_id IN (#{operable_ids.join(", ")})
+          UNION ALL
+          SELECT a.parent_id, a.child_id, root_node
+          FROM assignments a
+            INNER JOIN descendants d
+            ON a.parent_id = d.child_id
+          WHERE a.parent_id = d.child_id
+        )
+        SELECT d.root_node, crs.id
```
hstebbins-mdsol: Strong preference to alias the `crs.id` result to `crs_id` or something more specific in the SELECT, so the output column isn't just `id`.

hstebbins-mdsol:
I'm strongly leaning towards making this query more generic (perhaps even by just injecting the `active_policy_element_type` and separating the Dalton-specific context from the PM, especially considering this is a public repo.

lib/policy_machine_storage_adapters/active_record.rb
```diff
@@ -95,6 +95,179 @@ def self.persist_buffers!
...
+            INNER JOIN descendants d 
+            ON a.parent_id = d.child_id
+          WHERE a.parent_id = d.child_id
+        )
+        SELECT root_node, ra.operator_uri, crs.id
+        FROM descendants d
+          JOIN policy_element_associations AS pea
+          ON pea.object_attribute_id = d.child_id
+            OR pea.object_attribute_id = d.parent_id
+          JOIN policy_elements as ra
```
hstebbins-mdsol: I think I agree with @fpcyan that these aliases should be more semantic, at least for cases like this where the logic of the join depends on whether it's a role assignment, operation set, etc

DMcKinnon-mdsol:
@drozenberg should we close this now?

PR #142, opened 2018-06-28
[MCC-404025] Allow array parameters for finding records

hstebbins-mdsol:
This PR modifies the `find_all_of_type_#{pe_type}` method definition to allow for passing arrays as arguments to find records matching any of the elements in the array. 

Previously, this worked with only very specific conditions due to the logic that determines the `all` variable. I believe that you could pass arrays when finding based on a column attribute besides `name`. 

Note that this method still is extremely inefficient for querying on the `extra_attributes` column, as all searching is done in-memory. There is still a **lot** of potential performance improvements in this method, and it's used frequently enough that I think work in the future could drastically improve performance across many workflows. 

@mdsol/team04 - please review.

### DIFF
--- a/lib/policy_machine_storage_adapters/active_record.rb
+++ b/lib/policy_machine_storage_adapters/active_record.rb
@@ -775,17 +775,34 @@ module PolicyMachineStorageAdapter
       # If any condition is case-insensitive, the nodes need to be built
       # individually with Arel.
       if ignore_case
-        conditions.map do |k, v|
-          if ignore_case_applies?(ignore_case, k)
-            build_arel_insensitive(pe_class: pe_class, key: k, value: v)
-          else
-            build_arel_sensitive(pe_class: pe_class, key: k, value: v)
-          end
-        # Reduce Arel nodes into ActiveRecord_Relation
-        end.reduce(pe_class.where(nil)) { |rel, e| rel.where(e) }
+        build_ignore_case_relation(
+          pe_class: pe_class,
+          conditions: conditions,
+          ignore_case: ignore_case
+        )
       else
         # If all conditions are case-sensitive, a direct where call can be used.
         pe_class.where(conditions.to_h)
+      end
+    end
+
+    def build_ignore_case_relation(pe_class:, conditions:, ignore_case:)
+      # Conditions hash to array so it can be reduced
+      condition_array = conditions.to_a
+      # Initialize an "empty" relation
+      starting_relation = pe_class.where(nil)
+
+      # Reduce the conditions array to a single relation by building Arel nodes
+      condition_array.reduce(starting_relation) do |relation, condition_pair|
+        key, value = condition_pair
+
+        arel_node = if ignore_case_applies?(ignore_case, key)
+                      build_arel_insensitive(pe_class: pe_class, key: key, value: value)
+                    else
+                      build_arel_sensitive(pe_class: pe_class, key: key, value: value)
+                    end
+
+        relation.where(arel_node)
       end
     end
 
@@ -804,7 +821,7 @@ module PolicyMachineStorageAdapter
       # See: https://github.com/rails/arel/issues/368
       if value.is_a?(Array) && value.present?
         pe_class.arel_table[key].eq_any(value)
-      elsif value.is_a?(Array) && value.empty? 
+      elsif value.is_a?(Array) && value.empty?
         ::Arel::Nodes::SqlLiteral.new("(NULL)")
       else
         pe_class.arel_table[key].eq(value)
--- a/lib/policy_machine_storage_adapters/active_record.rb
+++ b/lib/policy_machine_storage_adapters/active_record.rb
@@ -476,156 +476,6 @@ module PolicyMachineStorageAdapter
       end
     end # End of POLICY_ELEMENT_TYPES iteration
 
-    # Given a policy element class and a set of conditions, returns an
-    # ActiveRecord_Relation with those conditions applied
-    def build_active_record_relation(pe_class:, conditions:, ignore_case:)
-      # If any condition is case-insensitive, the nodes need to be built
-      # individually with Arel.
-      if ignore_case
-        conditions.map do |k, v|
-          # Arel matches provides agnostic case insensitive sql for MySQL and
-          # Postgres. This should always evaluate to an ActiveRecord_Relation.
-          if ignore_case_applies?(ignore_case, k)
-            build_arel_insensitive(pe_class: pe_class, key: k, value: v)
-          else
-            build_arel_sensitive(pe_class: pe_class, key: k, value: v)
-          end
-        # Reduce Arel nodes into ActiveRecord_Relation
-        end.reduce(pe_class.where(nil)) { |rel, e| rel.where(e) }
-      else
-        # If all conditions are case-sensitive, a direct where call can be used.
-        pe_class.where(conditions.to_h)
-      end
-    end
-
-    # Build Arel nodes using case-insensitive matching
-    def build_arel_insensitive(pe_class:, key:, value:)
-      unless value.is_a?(Array)
-        pe_class.arel_table[key].matches(value)
-      else
-        pe_class.arel_table[key].matches_any(value)
-      end
-    end
-
-    # Build Arel nodes using case-sensitive equality checking
-    def build_arel_sensitive(pe_class:, key:, value:)
-      unless value.is_a?(Array)
-        pe_class.arel_table[key].eq(value)
-      else
-        unless value.empty?
-          pe_class.arel_table[key].eq_any(value)
-        else
-          # Arel blows up with empty array passed to eq_any
-          # See: https://github.com/rails/arel/issues/368
-          ::Arel::Nodes::SqlLiteral.new("(NULL)")
-        end
-      end
-    end
-
-    # Iterate over the input scope and return elements that match the extra
-    # attribute conditions
-    def filter_by_extra_attributes(scope:, extra_attribute_conditions:, pe_class:, ignore_case:)
-      unless extra_attribute_conditions.empty?
-        ids = scope.select do |pe|
-          pe_within_scope = true
-
-          extra_attribute_conditions.each do |key, value|
-            unless pe_matches_extra_attributes?(pe, key, value, ignore_case)
-              pe_within_scope = false
-            end
-          end
-
-          pe_within_scope
-        end.map(&:id)
-
-        pe_class.where(id: ids)
-      else
-        scope
-      end
-    end
-
-    # Check inclusion for each include condition using the Adapter
-    def filter_by_include_conditions(scope:, include_conditions:, klass:)
-      include_conditions.each do |key, value|
-        scope = Adapter.apply_include_condition(
-          scope: scope,
-          key: key,
-          value: value[:include],
-          klass: klass
-        )
-      end
-      scope
-    end
-
-    # Paginate the scope if options hash has pagination information
-    def paginate_scope(scope:, options:)
-      if options[:per_page]
-        page = options[:page] ? options[:page] : 1
-        scope = scope.order(:id).paginate(page: page, per_page: options[:per_page])
-      end
-      scope
-    end
-
-    # A value hash where the only key is :include is special.
-    # Note: If we start accepting literal hash values this may need to start checking the key's column type
-    def include_condition?(key, value)
-      value.respond_to?(:keys) && value.keys.map(&:to_sym) == [:include]
-    end
-
-    def class_for_type(pe_type)
-      @pe_type_class_hash ||= Hash.new { |h,k| h[k] = "PolicyMachineStorageAdapter::ActiveRecord::#{k.camelize}".constantize }
-      @pe_type_class_hash[pe_type]
-    end
-
-    # Check if the PolicyElement's extra_attributes column includes the passed
-    # key and value
-    def pe_matches_extra_attributes?(policy_element, key, value, ignore_case)
-      if policy_element.store_attributes
-        attr_value = policy_element.extra_attributes_hash[key]
-
-        if value.is_a?(Array)
-          value.any? do |v|
-            extra_attribute_match(
-              ignore_case: ignore_case,
-              key: key,
-              value_1: attr_value,
-              value_2: v
-            )
-          end
-        else
-          extra_attribute_match(
-            ignore_case: ignore_case,
-            key: key,
-            value_1: attr_value,
-            value_2: value
-          )
-        end
-      end
-    end
-
-    def extra_attribute_match(ignore_case:, key:, value_1:, value_2:)
-      if ignore_case_applies?(ignore_case, key) && value_1.is_a?(String) && value_2.is_a?(String)
-        value_1.downcase == value_2.downcase
-      else
-        value_1 == value_2
-      end
-    end
-
-    # Allow ignore_case to be a boolean, string, symbol, or array of symbols or strings
-    def ignore_case_applies?(ignore_case, key)
-      return false if key == 'policy_machine_uuid'
-
-      return true if ignore_case == true
-
-      # e.g. ignore_case = :name
-      return true if ignore_case.to_s == key
-
-      # e.g. ignore_case = [:name, :parent_uri]
-      return true if ignore_case.respond_to?(:any?) && ignore_case.any? { |k| k.to_s == key }
-
-      false
-    end
-
     ##
     # Assign src to dst in policy machine.
     # The two policy elements must be persisted policy elements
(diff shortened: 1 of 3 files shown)


lib/policy_machine_storage_adapters/active_record.rb
```diff
@@ -498,18 +526,47 @@ def pe_matches_extra_attributes?(policy_element, key, value, ignore_case)
...
+
+      return true if ignore_case == true
+
+      # i.e. ignore_case = :name, key = 'name'
+      return true if ignore_case.to_s == key
+
+      # i.e. ignore_case = [:name, :parent_uri], key = 'name'
+      return true if ignore_case.respond_to?(:any?) && ignore_case.any? { |k| k.to_s == key }
+
+      return false
```
DMcKinnon-mdsol: Don't need the `return` here.

lib/policy_machine_storage_adapters/active_record.rb
```diff
@@ -432,33 +432,75 @@ def operations
       end
 
       define_method("find_all_of_type_#{pe_type}") do |options = {}|
```
DMcKinnon-mdsol: This 70-line method is complex and difficult to review. I just checked, and nowhere in dalton do we use `find_all(..., ignore_case: true)`. We also don't appear to use `conditions` or `included_conditions`, if I'm understanding them correctly. It looks like we may have overdesigned this method to begin with, and now we're trying to maintain backwards compatibility for a bunch of functionality no one is using.

I'd be in favor of bumping the pm a major version and making the backwards incompatible change of simplifying this method to:

```ruby
define_method("find_all_of_type_#{pe_type}") do |options = {}|
  options[:unique_identifier] = options.delete(:uuid) if options[:uuid]
  class_for_type(pe_type).where(options.slice(*PolicyElement.column_names))
end
```

We _would_ have to maintain backwards compatibility for `extra_attributes`, so we'd need to tack that in somewhere as well, but I'd focus on simplifying that piece (instead of making it performant) because that bit already doesn't scale in the current implementation.

hstebbins-mdsol: `conditions` is essentially the hash that is used to filter column attributes. The main lookup of this method is in the first `all` variable assignment; the assignment maps over the hash and reduces it to an ActiveRecord_Relation object. 

Unfortunately it appears that we do use `included_conditions`. At first glance it seems the permissions controller uses it; I would need to check thoroughly to find other cases if they exist. 

Additionally it seems that `ignore_case` is set to `[:name]` by default in APE::Base. I haven't seen any other use for it except when `ignore_case` is set to false. I believe the reason for most of these explicit false settings were due to a previous bug in this method that sometimes didn't return results at all unless `ignore_case` was false. 

I definitely agree that this method could be trimmed. I support a major version bump to go with that cleanup, and removing a lot of the unused functionality. That being said, I think there are immediate benefits to be had with even these changes, namely being able to pass in arrays of UUIDs so we don't have to hit the db multiple times and map over UUIDs if we're doing mass lookups, like in CRS#create and #update. 

In my opinion, we should integrate these changes so we can start using the improved functionality with a minor version bump, then create a story to 1. simplify this method considerably and 2. clean up the usage in Dalton to take advantage of the improved method.

hstebbins-mdsol: @DMcKinnon-mdsol I've extracted the logic into several methods

lib/policy_machine_storage_adapters/active_record.rb
```diff
@@ -482,6 +482,97 @@ def operations
       end
     end # End of POLICY_ELEMENT_TYPES iteration
 
+    # Arel matches provides agnostic case insensitive sql for MySQL and
```
PoloPro: I would move this comment to line 487, where we actually see arel in use. For a comment on the method definition itself, maybe something like "Given a policy element class and a set of conditions, returns an ActiveRecord::Relation with those conditions applied".

spec/policy_machine_storage_adapters/active_record_spec.rb
```diff
@@ -26,21 +26,65 @@
...
+        expect(
+          policy_machine_storage_adapter.find_all_of_type_object(
+            unique_identifier: search_uuids
+          ).count
+        ).to eq(2)
+      end
+
+      context 'when case insensitive' do
+        it 'accepts an array parameter on a column attribute' do
+          colors = ['some_uuid1', 'some_uuid2']
```
PoloPro: Would prefer to see `colors` differ from the uuids of the objects, for clarity. `['burnt_umber', 'mauve']`

spec/policy_machine_storage_adapters/active_record_spec.rb
```diff
@@ -26,21 +26,65 @@
...
+      it 'allows uuid as a parameter' do
+        uuid = 'some_uuid'
+        policy_machine_storage_adapter.add_object(uuid, 'some_policy_machine_uuid')
 
+        expect(
+          policy_machine_storage_adapter.find_all_of_type_object(uuid: uuid).count
+        ).to eq(1)
+      end
+
+      context 'an extra attribute column has been added to the database' do
```
PoloPro: This is a little misleading, `extra_attributes` are faux column attributes by definition.

lib/policy_machine_storage_adapters/active_record.rb
```diff
@@ -432,47 +432,41 @@ def operations
...
-                  end
-                end.reduce(pe_class.where(nil)) { |rel, e| rel.where(e) }
-              else
-                pe_class.where(conditions.to_h)
-              end
-
-        include_conditions.each do |key, value|
-          all = Adapter.apply_include_condition(scope: all, key: key, value: value[:include], klass: class_for_type(pe_type))
-        end
+        results = build_active_record_relation(
```
DMcKinnon-mdsol: `results` seems pretty ambiguous. @hstebbins-mdsol what do you think about `active_record_relation` instead?

lib/policy_machine_storage_adapters/active_record.rb
```diff
@@ -432,47 +432,41 @@ def operations
       end
 
       define_method("find_all_of_type_#{pe_type}") do |options = {}|
+        options[:unique_identifier] = options.delete(:uuid) if options[:uuid]
         conditions = options.slice!(:per_page, :page, :ignore_case).stringify_keys
+        # Separate conditions on PolicyElement columns from conditions on the extra_attributes column
         extra_attribute_conditions = conditions.slice!(*PolicyElement.column_names)
         include_conditions, conditions = conditions.partition { |k,v| include_condition?(k,v) }
+        # Generated PolicyElement class
```
DMcKinnon-mdsol: Nitpick: we can increase readability a bit by adding some returns here and above.

lib/policy_machine_storage_adapters/active_record.rb
```diff
@@ -482,6 +476,96 @@ def operations
       end
     end # End of POLICY_ELEMENT_TYPES iteration
 
+    # Given a policy element class and a set of conditions, returns an
+    # ActiveRecord_Relation with those conditions applied
```
DMcKinnon-mdsol: * `ActiveRecord::Relation`

lib/policy_machine_storage_adapters/active_record.rb
```diff
@@ -482,6 +476,96 @@ def operations
...
+    def build_active_record_relation(pe_class:, conditions:, ignore_case:)
+      # If any condition is case-insensitive, the nodes need to be built
+      # individually with Arel.
+      if ignore_case
+        conditions.map do |k, v|
+          # Arel matches provides agnostic case insensitive sql for MySQL and
```
DMcKinnon-mdsol: Looks like "matches" should be removed here.

lib/policy_machine_storage_adapters/active_record.rb
```diff
@@ -482,6 +476,96 @@ def operations
...
+        end.reduce(pe_class.where(nil)) { |rel, e| rel.where(e) }
+      else
+        # If all conditions are case-sensitive, a direct where call can be used.
+        pe_class.where(conditions.to_h)
+      end
+    end
+
+    # Build Arel nodes using case-insensitive matching
+    def build_arel_insensitive(pe_class:, key:, value:)
+      unless value.is_a?(Array)
```
DMcKinnon-mdsol: https://github.com/rubocop-hq/ruby-style-guide#no-else-with-unless

DMcKinnon-mdsol: Could also make this method
```ruby
def build_arel_insensitive(pe_class:, key:, value:)
  pe_class.arel_table[key].matches_any([value].flatten(1))
end
```
...not sure if that's clearer, though.

lib/policy_machine_storage_adapters/active_record.rb
```diff
@@ -482,6 +476,96 @@ def operations
...
+        pe_class.arel_table[key].matches(value)
+      else
+        pe_class.arel_table[key].matches_any(value)
+      end
+    end
+
+    # Build Arel nodes using case-sensitive equality checking
+    def build_arel_sensitive(pe_class:, key:, value:)
```
DMcKinnon-mdsol: I'd prefer to avoid nested if/unless whenever possible. How about:
```ruby
def build_arel_sensitive(pe_class:, key:, value:)
  if !value.is_a?(Array)
    pe_class.arel_table[key].eq(value)
  elsif value.present?
    pe_class.arel_table[key].eq_any(value)
  else
    # Arel blows up with empty array passed to eq_any
    # See: https://github.com/rails/arel/issues/368
    ::Arel::Nodes::SqlLiteral.new("(NULL)")
  end
end
```

lib/policy_machine_storage_adapters/active_record.rb
```diff
@@ -482,6 +476,96 @@ def operations
...
+          ::Arel::Nodes::SqlLiteral.new("(NULL)")
+        end
+      end
+    end
+
+    # Iterate over the input scope and return elements that match the extra
+    # attribute conditions
+    def filter_by_extra_attributes(scope:, extra_attribute_conditions:, pe_class:, ignore_case:)
+      unless extra_attribute_conditions.empty?
+        ids = scope.select do |pe|
```
DMcKinnon-mdsol: Looks like we're iterating over this scope twice – once for the select and then once again for the map at the end. If we use a reduce here instead then we can cut down our number of iterations over the scope to 1.

lib/policy_machine_storage_adapters/active_record.rb
```diff
@@ -482,6 +476,96 @@ def operations
...
+        end.map(&:id)
+
+        pe_class.where(id: ids)
+      else
+        scope
+      end
+    end
+
+    # Check inclusion for each include condition using the Adapter
+    def filter_by_include_conditions(scope:, include_conditions:, klass:)
```
DMcKinnon-mdsol: Should all these helper methods be private?

lib/policy_machine_storage_adapters/active_record.rb
```diff
@@ -482,6 +476,96 @@ def operations
...
+          klass: klass
+        )
+      end
+      scope
+    end
+
+    # Paginate the scope if options hash has pagination information
+    def paginate_scope(scope:, options:)
+      if options[:per_page]
+        page = options[:page] ? options[:page] : 1
```
DMcKinnon-mdsol:
```ruby
page = options[:page] || 1
```

lib/policy_machine_storage_adapters/active_record.rb
```diff
@@ -493,23 +577,53 @@ def class_for_type(pe_type)
...
+        value_1 == value_2
+      end
+    end
+
     # Allow ignore_case to be a boolean, string, symbol, or array of symbols or strings
     def ignore_case_applies?(ignore_case, key)
       return false if key == 'policy_machine_uuid'
-      ignore_case == true || ignore_case.to_s == key || ( ignore_case.respond_to?(:any?) && ignore_case.any? { |k| k.to_s == key } )
+
+      return true if ignore_case == true
```
DMcKinnon-mdsol: Can we just do the following?
```ruby
return true if ignore_case
```

DMcKinnon-mdsol: Ah, just saw the lines below. Retracted.

spec/policy_machine_storage_adapters/active_record_spec.rb
```diff
@@ -26,21 +26,65 @@
     let(:policy_machine_storage_adapter) { described_class.new }
 
     describe 'find_all_of_type' do
+      it 'accepts an array parameter on a column attribute' do
+        search_uuids = ['some_uuid1', 'some_uuid2']
```
DMcKinnon-mdsol: Do we need to set the uuids specifically? Can we just create two generic objects and search for the uuids of the things we created?

lib/policy_machine_storage_adapters/active_record.rb
hstebbins-mdsol: excellent idea

lib/policy_machine_storage_adapters/active_record.rb
hstebbins-mdsol: Weirdly enough, the class that is returned is `ActiveRecord_Relation`

lib/policy_machine_storage_adapters/active_record.rb
hstebbins-mdsol: That's a tricky call. I think I would prefer being explicit about what we're expecting `value` to possibly be... I think it would be better to err on the side of being verbose over concise (just for straightforwardness) in these two methods, but I'm not 100% attached to that.

lib/policy_machine_storage_adapters/active_record.rb
hstebbins-mdsol: Good idea, but I made a slightly different decision tree. Take a look and see if it makes sense to you.

lib/policy_machine_storage_adapters/active_record.rb
hstebbins-mdsol: If only the world were that simple!

lib/policy_machine_storage_adapters/active_record.rb
hstebbins-mdsol: Excellent catch!

lib/policy_machine_storage_adapters/active_record.rb
DMcKinnon-mdsol: Oh lol okay.

lib/policy_machine_storage_adapters/active_record.rb
DMcKinnon-mdsol: That works for me 👍

lib/policy_machine_storage_adapters/active_record.rb
hstebbins-mdsol: Since directly returning a null Arel node is so rare, weird, and context-specific, I wanted to be as obvious about what was happening as possible. I'm fine with changing it, though.

hstebbins-mdsol:
@DMcKinnon-mdsol I noticed that it was iterating over `conditions` in `build_active_record_relation` twice as well, so I broke that up and used `reduce` as well. Just FYI in case you want to take a look.

spec/policy_machine_storage_adapters/active_record_spec.rb
```diff
@@ -26,25 +26,71 @@
...
+        ).to eq(2)
+      end
 
-      it 'warns once when filtering on an extra attribute' do
-        expect(Warn).to receive(:warn).once
-        2.times do
-          expect(policy_machine_storage_adapter.find_all_of_type_user(foo: 'bar')).to be_empty
+      context 'when case insensitive' do
+        it 'accepts an array parameter on a column attribute' do
+          pm_uuid = SecureRandom.uuid
```
drozenberg: I would just use a let block up after the describe since this is referenced a lot`let(:pm_uuid) { SecureRandom.uuid }`

spec/policy_machine_storage_adapters/active_record_spec.rb
hstebbins-mdsol: Being able to create generic objects would be ideal, but it's a change that would be a bit outside the scope of this PR I think. I changed it to at least generate uuids instead of having hardcoded strings within the specs I touched.

lib/policy_machine_storage_adapters/active_record.rb
fpcyan: `ignore_case: true` is default, so Dalton uses it almost _everywhere_

lib/policy_machine_storage_adapters/active_record.rb
```diff
@@ -798,6 +769,168 @@ def accessible_objects(user_or_attribute, operation, options = {})
...
+      else
+        ids = scope.reduce([]) do |memo, policy_element|
+          pe_within_scope = extra_attribute_conditions.all? do |key, value|
+            pe_matches_extra_attributes?(policy_element, key, value, ignore_case)
+          end
+
+          memo.push(policy_element.id) if pe_within_scope
+          memo
+        end
+
```
DMcKinnon-mdsol: Extra space snuck in here

lib/policy_machine_storage_adapters/active_record.rb
```diff
@@ -798,6 +769,168 @@ def accessible_objects(user_or_attribute, operation, options = {})
...
+    def filter_by_extra_attributes(scope:, extra_attribute_conditions:, pe_class:, ignore_case:)
+      if extra_attribute_conditions.empty?
+        scope
+      else
+        ids = scope.reduce([]) do |memo, policy_element|
+          pe_within_scope = extra_attribute_conditions.all? do |key, value|
+            pe_matches_extra_attributes?(policy_element, key, value, ignore_case)
+          end
+
+          memo.push(policy_element.id) if pe_within_scope
```
DMcKinnon-mdsol: We could use a ternary operator if that's your kind of thing. I'm also fine with this, though.

lib/policy_machine_storage_adapters/active_record.rb
hstebbins-mdsol: What do you mean @fpcyan ? From what I've seen, the default value of `ignore_case` that is passed into this method is `[:name]`, but I might have missed a default somewhere else.

lib/policy_machine_storage_adapters/active_record.rb
hstebbins-mdsol: I'm not sure what you mean - I don't really see how a ternary could be used here. Mind sharing more detail?

DMcKinnon-mdsol: Like
```ruby
pe_within_scope ? memo.push(policy_element.id) : memo
```

DMcKinnon-mdsol:
Merging in 30 minutes unless objections.

PR #151, opened 2018-08-09
[MCC-409765] Update PM to utilize soft deletion functionality in privilege calls 

hstebbins-mdsol:
This PR updates the Police Machine to utilize soft deletion functionality in ActiveRecord privilege calls. 

It modifies: 
1. The `is_privilege?` definition on the ActiveRecord adapter by adding default behavior to ignore soft deleted data, with an optional hash parameter that can change this behavior by passing `true` to the `:use_soft_deleted_data` key 
2. The PM test app setup in relation to ActiveRecord adapter testing by adding a deleted_at column
3. Specs for `is_privilege?` calls and `is_privilege_without_prohibitions?` calls on the ActiveRecord adapter

@mdsol/team04 - please review.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,10 +1,13 @@
 # Changelog
 
+## 1.8.1
+* Add soft deletion support to the ActiveRecord storage adapter.
+
 ## 1.8.0
 * Add accessible_ancestor_objects method to ActiveRecord storage adapter.
 
 ## 1.7.4
-* Re-expose the 'class_for_type' method to the public interface. 
+* Re-expose the 'class_for_type' method to the public interface.
 
 ## 1.7.3
 * Improve find_all_of_type_* functionality. This allows for properly passing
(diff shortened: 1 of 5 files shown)


hstebbins-mdsol:
Note that I will have to bump the PM version if we decide to accept this methodology (this is more a note to myself)

DMcKinnon-mdsol:
Why did you decide not to go with this approach?

hstebbins-mdsol:
This added a bit of Dalton-specific logic to the PM (e.g. a `deleted_at` column), and it required a _lot_ of explicit changes. 

After discussing it with Connor and Wes, I think a better method would be to add a configuration option to the PM for injecting a default scope to PolicyElement. This minimizes Dalton-specific logic in the PM as well as allows more flexible specs. You can see the specific details of the proposed change in the Soft Deleting ADR PR in Dalton.

PR #152, opened 2018-08-13
[MCC-409765]  Add PolicyElement default scoping

hstebbins-mdsol:
This PR adds the ability to inject a default scope to PolicyElement into the Policy Machine.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+## 1.9.0
+* Add optional configuration object to the Policy Machine
+* Implement a default scope for Policy Elements
+
 ## 1.8.0
 * Add accessible_ancestor_objects method to ActiveRecord storage adapter.
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,8 +1,8 @@
 # Changelog
 
 ## 1.9.0
-* Add optional configuration object to the Policy Machine
-* Implement a default scope for Policy Elements
+* Add optional configuration object to the Policy Machine.
+* Implement a default scope for Policy Elements.
 
 ## 1.8.0
 * Add accessible_ancestor_objects method to ActiveRecord storage adapter.
--- a/lib/policy_machine_storage_adapters/active_record.rb
+++ b/lib/policy_machine_storage_adapters/active_record.rb
@@ -114,6 +114,12 @@ module PolicyMachineStorageAdapter
 
       active_record_serialize :extra_attributes, JSON
 
+      default_scope {
+        if PolicyMachine.configuration.policy_element_default_scope
+          where(PolicyMachine.configuration.policy_element_default_scope => nil)
+        end
+      }
+
       def method_missing(meth, *args, &block)
         store_attributes
         if respond_to?(meth)
@@ -792,7 +798,7 @@ module PolicyMachineStorageAdapter
       # Fetch all of the PEAs using the given UA or its descendants
       user_attribute_ids = user_or_attribute.descendants.pluck(:id) | [user_or_attribute.id]
       associations = PolicyElementAssociation.where(user_attribute_id: user_attribute_ids)
-      operation_set_ids = associations.pluck(:operation_set_id)
+      operation_set_ids = OperationSet.where(id: associations.select(:operation_set_id)).pluck(:id)
 
       # Narrow the list of PEAs to just those containing the specified operation
       operation_id = operation.try(:unique_identifier) || operation.to_s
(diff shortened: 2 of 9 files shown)


lib/policy_machine_storage_adapters/active_record.rb
```diff
@@ -1000,7 +1006,9 @@ def accessible_operations(user_or_attribute, object_or_attribute, operation_id =
         prms = { type: PolicyMachineStorageAdapter::ActiveRecord::Operation.to_s }
         prms.merge!(unique_identifier: operation_id) if operation_id
 
-        Assignment.descendants_of(associations.map(&:operation_set)).where(prms)
+        operation_sets = OperationSet.where(id: associations.select(:operation_set_id))
```
DMcKinnon-mdsol: Looks like you should be able to use `pluck` instead of `select` here.

DMcKinnon-mdsol:
Interesting, Harry! What was the impetus for doing this now?

Also, mind opening a dalton PR using this branch?

lib/policy_machine_storage_adapters/active_record.rb
```diff
@@ -792,7 +798,7 @@ def accessible_ancestor_objects(user_or_attribute, operation, root_object, optio
       # Fetch all of the PEAs using the given UA or its descendants
       user_attribute_ids = user_or_attribute.descendants.pluck(:id) | [user_or_attribute.id]
       associations = PolicyElementAssociation.where(user_attribute_id: user_attribute_ids)
-      operation_set_ids = associations.pluck(:operation_set_id)
+      operation_set_ids = OperationSet.where(id: associations.select(:operation_set_id)).pluck(:id)
```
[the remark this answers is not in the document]
hstebbins-mdsol: So, if a record is soft-deleted, that soft deletion will currently _not_ affect related records in the `associations`, `assignments`, etc. 

This fixes the issue by explicitly trying to find the Operation Set that is implied by the existence of the association, instead of relying on the implicit existence from the association. 

Interestingly enough, ActiveRecord/Arel is actually intelligent enough to _not_ create another SQL call! Instead, it will create a subquery!

Check this out: 
```sql
[9] pry(#<PolicyMachineStorageAdapter::ActiveRecord>)> puts OperationSet.where(id: associations.select(:operation_set_id)).to_sql
SELECT "policy_elements".* FROM "policy_elements" WHERE "policy_elements"."type" IN ('PolicyMachineStorageAdapter::ActiveRecord::OperationSet') AND "policy_elements"."id" IN (SELECT "policy_element_associations"."operation_set_id" FROM "policy_element_associations" WHERE "policy_element_associations"."user_attribute_id" IN (2, 1) AND "policy_element_associations"."object_attribute_id" IN (5, 6))
=> nil
```

A prettier version of the output: 
```sql
SELECT "policy_elements".*
FROM "policy_elements"
WHERE "policy_elements"."type" IN ('PolicyMachineStorageAdapter::ActiveRecord::OperationSet')
AND "policy_elements"."id" IN (
  SELECT "policy_element_associations"."operation_set_id"
  FROM "policy_element_associations"
  WHERE "policy_element_associations"."user_attribute_id" IN (2, 1)
    AND "policy_element_associations"."object_attribute_id" IN (5, 6)
)


```

lib/policy_machine_storage_adapters/active_record.rb
hstebbins-mdsol: See comment above - `pluck` will actually load the IDs, while we just want the _relation_ used to build the query, similar to above.

hstebbins-mdsol:
This PR matches up with the [soft deleting operation sets](https://github.com/mdsol/dalton/pull/2647) PR. 😄 

Handling soft deleted data in Dalton right now is quite messy, and involves manually rejecting soft deleted data as it's collected in different workflows. Instead of trying to do something similar in the PM, it makes sense to add a default scoping ability to the PM to simply ignore soft deleted data in the data layer entirely. It ends up being a lot cleaner - although, due to custom SQL in the Postgres adapter, it won't be as simple as a single scope, unfortunately. Still, it's much cleaner overall. 

Here, we can now do something like:
```ruby 
unless USE_SOFT_DELETED_DATA
  PolicyMachine.configure do |config|
    config.policy_element_default_scope = { where(deleted_at: nil) }
  end
end
```
during Dalton's initialization. I haven't finalized the exact interface yet, so consider above to be pseudocode, but it illustrates what I'd like ideally.

hstebbins-mdsol:
@mdsol/team04 - this PR is now ready for review.

CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # Changelog
 
+## 1.9.0
+* Add optional configuration object to the Policy Machine
```
DMcKinnon-mdsol: Nitpick: periods, please.

DMcKinnon-mdsol:
Any additional comments here @mdsol/team-04? If not I'll merge at 1pm EST.

hstebbins-mdsol:
I actually think we should hold off merging this until tomorrow, to allow me to do some performance testing of the new changes. I don't expect significant degradations, but if there are, I don't think it would be worth it to merge.

DMcKinnon-mdsol:
Closing pending a re-discussion for performance reasons.

PR #153, opened 2018-08-23
[MCC-229209] Upgrade the PM to support Rails 5.2

hstebbins-mdsol:
This PR upgrades the Policy Machine to Rails 5.2.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,7 @@
 # Changelog
 
 ## 2.0.0
-* Upgrade the Policy Machine to support Rails 5.2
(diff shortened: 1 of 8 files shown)


DMcKinnon-mdsol:
Is it still going to be backwards compatible with Rails 4?

hstebbins-mdsol:
I doubt it, primarily due to the change in how JSONB serialization is handled, but I'll give it a shot and update here.

hstebbins-mdsol:
Thankfully, I was wrong! It appears to still be compatible.

DMcKinnon-mdsol:
Merging in 2hrs unless objections @mdsol/team-04.

PR #155, opened 2018-09-10
[MCC-224756] Privileges scoped by role

hstebbins-mdsol:
This PR updates several methods in the PM and adds a new one. These additions allow for consumers to scope various privilege-related things with a given user attribute context.

The new method is `is_privilege_in_role?`, which functions nearly the same as `is_privilege?` except it only returns true if the privilege is granted by the given user attribute or its descendants. 

The updates methods are `accessible_objects`, `accessible_ancestor_objects`, `accessible_operations`, and `scoped_privileges`. These now allow for an optional `:user_attribute_scope` entry in the options hash. These methods will only return results given by the specified user attribute scope or its descendants.

### DIFF
--- a/lib/policy_machine.rb
+++ b/lib/policy_machine.rb
@@ -282,7 +282,8 @@ class PolicyMachine
   end
 
   def is_privilege_in_role?(user_or_attribute, operation, object_or_attribute, user_attribute_scope, options = {})
-    is_privilege_in_role_ignoring_prohibitions?(user_or_attribute, operation, object_or_attribute, user_attribute_scope) && (options[:ignore_prohibitions] || !is_privilege_in_role_ignoring_prohibitions?(user_or_attribute, PM::Prohibition.on(operation), object_or_attribute, user_attribute_scope))
+    # Prohibitions shouldn't ever be scoped to role, so use the normal privilege check for the prohibition
+    is_privilege_in_role_ignoring_prohibitions?(user_or_attribute, operation, object_or_attribute, user_attribute_scope) && (options[:ignore_prohibitions] || !is_privilege_ignoring_prohibitions?(user_or_attribute, PM::Prohibition.on(operation), object_or_attribute))
   end
 
   def is_privilege_in_role_ignoring_prohibitions?(user_or_attribute, operation, object_or_attribute, user_attribute_scope)
--- a/lib/policy_machine.rb
+++ b/lib/policy_machine.rb
@@ -212,6 +212,9 @@ class PolicyMachine
     end
 
     if user_attribute_scope
+      # TODO: Surface user attribute origin in scoped_privileges_and_prohibitions
+      # so this call is unnecessary and the user attribute scoping can be done
+      # in-memory while maintaining the full set of prohibitions
       scoped_privs = scoped_privileges_and_prohibitions(
         user_or_attribute,
         object_or_attribute,
--- a/lib/policy_machine.rb
+++ b/lib/policy_machine.rb
@@ -115,12 +115,12 @@ class PolicyMachine
       return policy_machine_storage_adapter.is_privilege?(*privilege)
     end
 
-    if (user_attribute_scope = options[:user_attribute_scope]) && policy_machine_storage_adapter.respond_to?(:is_privilege_in_role?)
+    if (user_attribute_scope = options[:user_attribute_scope]) && policy_machine_storage_adapter.respond_to?(:is_privilege_via_attribute?)
       privilege = [user_or_attribute, operation, object_or_attribute, user_attribute_scope].map do |obj|
         obj.respond_to?(:stored_pe) ? obj.stored_pe : obj
       end
 
-      return policy_machine_storage_adapter.is_privilege_in_role?(*privilege)
+      return policy_machine_storage_adapter.is_privilege_via_attribute?(*privilege)
     end
 
     unless operation.is_a?(PM::Operation)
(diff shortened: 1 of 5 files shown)


lib/policy_machine.rb
```diff
@@ -348,6 +380,20 @@ def bulk_persist
...
 
+
+  def scoped_privileges_and_prohibitions(user_or_attribute, object_or_attribute, options)
+    if policy_machine_storage_adapter.respond_to?(:scoped_privileges)
+      policy_machine_storage_adapter.scoped_privileges(user_or_attribute.stored_pe, object_or_attribute.stored_pe, options).map do |op|
+        operation = PM::Operation.convert_stored_pe_to_pe(op, policy_machine_storage_adapter, PM::Operation)
+        [user_or_attribute, operation, object_or_attribute]
+      end
+    else
+      operations.grep(->operation{is_privilege_ignoring_prohibitions?(user_or_attribute, operation, object_or_attribute)}) do |op|
```
PoloPro: So this selects the operations that satisfy the privilege triplet, ignoring prohibitions. I'm not sure passing a proc to Enumerable's `grep` option is the clearest way to do that.

```ruby
operations.each_with_object([]) do |op, memo|
  if is_privilege_ignoring_prohibitions?(user_or_attribute, op, object_or_attribute)
    memo << [user_or_attribute, op, object_or_attribute]
  end
end
```

lib/policy_machine.rb
```diff
@@ -115,6 +115,11 @@ def is_privilege_ignoring_prohibitions?(user_or_attribute, operation, object_or_
       return policy_machine_storage_adapter.is_privilege?(*privilege)
     end
 
+    if (user_attribute_scope = options[:user_attribute_scope]) && policy_machine_storage_adapter.respond_to?(:is_privilege_in_role?)
+      privilege = [user_or_attribute, operation, object_or_attribute, user_attribute_scope].map { |obj| obj.respond_to?(:stored_pe) ? obj.stored_pe : obj }
```
PoloPro: Line's pretty long

lib/policy_machine_storage_adapters/active_record.rb
```diff
@@ -709,17 +709,37 @@ def is_privilege?(user_or_attribute, operation, object_or_attribute)
       end
     end
 
+    # Returns true if the user has the operation on the object, and that operation is granted by
+    # the user attribute scope or its descendants
+    def is_privilege_in_role?(user_or_attribute, operation, object_or_attribute, user_attribute_scope)
```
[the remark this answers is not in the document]
hstebbins-mdsol: For some reason this comment is marked as hidden in Github - do you still want this addressed?

lib/policy_machine_storage_adapters/active_record.rb
```diff
@@ -807,7 +849,7 @@ def associations_filtered_by_operation(associations, operation)
     # Builds an array of PolicyElement objects within the scope of a given
     # array of associations
     def build_accessible_object_scope(associations, options = {})
-      permitting_oas = PolicyElement.where(id: associations.pluck(:object_attribute_id))
+      permitting_oas = PolicyElement.where(id: associations.select(:object_attribute_id))
```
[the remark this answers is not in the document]
hstebbins-mdsol: it actually prevents an extra query, only possible because this is vanilla activerecord and those guys are really smart: 
```ruby
> PolicyElement.where(id: associations.select(:object_attribute_id)).to_sql
# => "SELECT \"policy_elements\".* FROM \"policy_elements\" WHERE \"policy_elements\".\"id\" IN (SELECT \"policy_element_associations\".\"object_attribute_id\" FROM \"policy_element_associations\" WHERE \"policy_element_associations\".\"user_attribute_id\" IN (2, 4) AND \"policy_element_associations\".\"operation_set_id\" = 17)"
> PolicyElement.where(id: associations.pluck(:object_attribute_id)).to_sql
# => "SELECT \"policy_elements\".* FROM \"policy_elements\" WHERE \"policy_elements\".\"id\" = 11"
```
note how `pluck` retrieves the id explicitly, meaning it loads it when it's resolving the entire line, while `select` generates into sql

lib/policy_machine_storage_adapters/active_record.rb
```diff
@@ -985,11 +1027,16 @@ def prohibition_for(operation)
...
 
-    def accessible_operations(user_or_attribute, object_or_attribute, operation_id = nil)
+    def accessible_operations(user_or_attribute, object_or_attribute, operation_id = nil, user_attribute_scope = nil)
       transaction_without_mergejoin do
         user_attribute_ids = Assignment.descendants_of(user_or_attribute).pluck(:id) | [user_or_attribute.id]
         object_attribute_ids = Assignment.descendants_of(object_or_attribute).pluck(:id) | [object_or_attribute.id]
 
+        if user_attribute_scope
+          scoped_user_attribute_ids = Assignment.descendants_of(user_attribute_scope).pluck(:id) | [user_attribute_scope.id]
+          user_attribute_ids = user_attribute_ids & scoped_user_attribute_ids
```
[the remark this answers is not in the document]
hstebbins-mdsol: i love it, done

spec/policy_machine_storage_adapters/active_record_spec.rb
```diff
@@ -25,6 +25,282 @@
...
+      describe 'scoped_privileges' do
+        context 'when the user has access via an operation set' do
+          it 'returns all the privileges granted by that role' do
+            expect(
+              candy_pm.scoped_privileges(
+                frank,
+                brie,
+                user_attribute_scope: cheese_engineer
+              )
+            ).to match_array([[frank, create, brie], [frank, taste, brie]])
```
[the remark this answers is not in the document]
hstebbins-mdsol: perfect, thanks!

lib/policy_machine.rb
hstebbins-mdsol: oh god trust me, I agree. This was original code I refactored from elsewhere, I don't like it but I don't think this PR is where I should change it

lib/policy_machine_storage_adapters/active_record.rb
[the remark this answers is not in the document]
PoloPro: Yes, please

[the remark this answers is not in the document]
hstebbins-mdsol: it's an excellent point, and frankly it probably should be more like `is_privilege_via_attribute` or something, since roles aren't really a PM thing anyways

lib/policy_machine_storage_adapters/active_record.rb
```diff
@@ -709,17 +709,37 @@ def is_privilege?(user_or_attribute, operation, object_or_attribute)
...
+    # Returns true if the user has the operation on the object, and that operation is granted by
+    # the user attribute scope or its descendants
+    def is_privilege_via_attribute?(user_or_attribute, operation, object_or_attribute, user_attribute_scope)
+      policy_classes_containing_object = policy_classes_for_object_attribute(object_or_attribute)
+      operation_id = operation.try(:unique_identifier) || operation.to_s
+
+      if policy_classes_containing_object.size < 2
+        !accessible_operations(user_or_attribute, object_or_attribute, operation_id, user_attribute_scope).empty?
+      else
+        policy_classes_containing_object.all? do |policy_class|
```
CGDS: could probably clean up the `arr.all? { |el| !el.empty?}` with something like `arr.any? { |el| el.empty?`, but it doesn't look like policy class is actually used in the block.  Did we add tests for multiple policy classes?

lib/policy_machine_storage_adapters/active_record.rb
```diff
@@ -709,17 +709,37 @@ def is_privilege?(user_or_attribute, operation, object_or_attribute)
...
+    # descendants
     def scoped_privileges(user_or_attribute, object_or_attribute, options = {})
       policy_classes_containing_object = policy_classes_for_object_attribute(object_or_attribute)
 
+      user_attribute_scope = options[:user_attribute_scope]
+
       operations =
         if policy_classes_containing_object.size < 2
-          accessible_operations(user_or_attribute, object_or_attribute)
+          accessible_operations(user_or_attribute, object_or_attribute, nil, user_attribute_scope)
```
CGDS: not a huge fan of needing to pass these `nil`s as args, but updating is probably out of scope for this pr

lib/policy_machine.rb
```diff
@@ -194,22 +202,37 @@ def privileges
...
-      policy_machine_storage_adapter.scoped_privileges(user_or_attribute.stored_pe, object_or_attribute.stored_pe, options).map do |op|
-        operation = PM::Operation.convert_stored_pe_to_pe(op, policy_machine_storage_adapter, PM::Operation)
-        [user_or_attribute, operation, object_or_attribute]
-      end
-    else
-      operations.grep(->operation{is_privilege_ignoring_prohibitions?(user_or_attribute, operation, object_or_attribute)}) do |op|
-        [user_or_attribute, op, object_or_attribute]
+    # Duplicate so passed hash doesn't get modified
+    options = options.dup
+    user_attribute_scope = options.delete(:user_attribute_scope)
```
CGDS: if we have activesupport in scope, this and the `dup` could be replaced with `except`, whether or not we _should_ have activesupport in scope is another question, though

hstebbins-mdsol: This workflow is a bit weird; `:user_attribute_scope` actually _must_ be removed from the options hash in order to get the full list of prohibitions, which is then filtered down with a second call that gets the privileges within the scope. It's not ideal, but until the TODO is completed (surfacing the user attributes that provide a given privilege) then it's necessary, I think

hstebbins-mdsol:
I misunderstood the requirements in this PR. Filtering based on user attributes does not need to cascade, so the work necessary is much simpler and lower touch. I'm closing this PR in favor of the approach in: https://github.com/mdsol/the_policy_machine/pull/157

PR #157, opened 2018-09-20
[MCC-224756] Filter privileges by user attribute

hstebbins-mdsol:
This PR updates the Policy Machine to allow filtering privilege APIs so they only return privileges (or whether a privilege exists) if it is derived via a user attribute that passes the filter.

### DIFF
--- a/lib/policy_machine.rb
+++ b/lib/policy_machine.rb
@@ -103,17 +103,7 @@ class PolicyMachine
   ##
   # Check the privilege with filters without checking for prohibitions.
   def is_filtered_privilege_ignoring_prohibitions?(user_or_attribute, operation, object_or_attribute, filters: {}, options: {})
-    unless user_or_attribute.is_a?(PM::User) || user_or_attribute.is_a?(PM::UserAttribute)
-      raise(ArgumentError, "user_attribute_pe must be a User or UserAttribute.")
-    end
-
-    unless [PM::Operation, Symbol, String].any? { |allowed_type| operation.is_a?(allowed_type) }
-      raise(ArgumentError, "operation must be an Operation, Symbol, or String.")
-    end
-
-    unless object_or_attribute.is_a?(PM::Object) || object_or_attribute.is_a?(PM::ObjectAttribute)
-      raise(ArgumentError, "object_or_attribute must either be an Object or ObjectAttribute.")
-    end
+    assert_privilege_parameters!(user_or_attribute, operation, object_or_attribute)
 
     if policy_machine_storage_adapter.respond_to?(:is_filtered_privilege?)
       policy_machine_storage_adapter.is_filtered_privilege?(user_or_attribute, operation, object_or_attribute, filters: filters, options: options)
@@ -126,17 +116,7 @@ class PolicyMachine
   # Check the privilege without checking for prohibitions. May be called directly but is also used in is_privilege?
   #
   def is_privilege_ignoring_prohibitions?(user_or_attribute, operation, object_or_attribute, options = {})
-    unless user_or_attribute.is_a?(PM::User) || user_or_attribute.is_a?(PM::UserAttribute)
-      raise(ArgumentError, "user_attribute_pe must be a User or UserAttribute.")
-    end
-
-    unless [PM::Operation, Symbol, String].any? { |allowed_type| operation.is_a?(allowed_type) }
-      raise(ArgumentError, "operation must be an Operation, Symbol, or String.")
-    end
-
-    unless object_or_attribute.is_a?(PM::Object) || object_or_attribute.is_a?(PM::ObjectAttribute)
-      raise(ArgumentError, "object_or_attribute must either be an Object or ObjectAttribute.")
-    end
+    assert_privilege_parameters!(user_or_attribute, operation, object_or_attribute)
 
     if options.empty? && policy_machine_storage_adapter.respond_to?(:is_privilege?)
       privilege = [user_or_attribute, operation, object_or_attribute].map { |obj| obj.respond_to?(:stored_pe) ? obj.stored_pe : obj }
@@ -386,6 +366,21 @@ class PolicyMachine
   end
 
   private
+
+  # Raise unless the given parameters are valid types
+  def assert_privilege_parameters!(user_or_attribute, operation, object_or_attribute)
+    unless user_or_attribute.is_a?(PM::User) || user_or_attribute.is_a?(PM::UserAttribute)
+      raise(ArgumentError, "user_attribute_pe must be a User or UserAttribute.")
+    end
+
+    unless [PM::Operation, Symbol, String].any? { |allowed_type| operation.is_a?(allowed_type) }
+      raise(ArgumentError, "operation must be an Operation, Symbol, or String.")
+    end
+
+    unless object_or_attribute.is_a?(PM::Object) || object_or_attribute.is_a?(PM::ObjectAttribute)
+      raise(ArgumentError, "object_or_attribute must either be an Object or ObjectAttribute.")
+    end
+  end
 
   # Retrieves all privileges and prohibitions for the given user or attribute on the object or attribute scope
   def get_all_scoped_privileges_and_prohibitions(user_or_attribute, object_or_attribute, options = {})
--- a/lib/policy_machine.rb
+++ b/lib/policy_machine.rb
@@ -97,12 +97,12 @@ class PolicyMachine
   ##
   # Can we derive a privilege given a filtered set of user attributes?
   def is_privilege_with_filters?(user_or_attribute, operation, object_or_attribute, filters: {}, options: {})
-    is_filtered_privilege_ignoring_prohibitions?(user_or_attribute, operation, object_or_attribute, filters: filters, options: options) && (options[:ignore_prohibitions] || !is_privilege_ignoring_prohibitions?(user_or_attribute, PM::Prohibition.on(operation), object_or_attribute, options))
+    is_privilege_ignoring_prohibitions_with_filters?(user_or_attribute, operation, object_or_attribute, filters: filters, options: options) && (options[:ignore_prohibitions] || !is_privilege_ignoring_prohibitions?(user_or_attribute, PM::Prohibition.on(operation), object_or_attribute, options))
   end
 
   ##
   # Check the privilege with filters without checking for prohibitions.
-  def is_filtered_privilege_ignoring_prohibitions?(user_or_attribute, operation, object_or_attribute, filters: {}, options: {})
+  def is_privilege_ignoring_prohibitions_with_filters?(user_or_attribute, operation, object_or_attribute, filters: {}, options: {})
     assert_privilege_parameters!(user_or_attribute, operation, object_or_attribute)
 
     if policy_machine_storage_adapter.respond_to?(:is_filtered_privilege?)
--- a/lib/policy_machine.rb
+++ b/lib/policy_machine.rb
@@ -204,18 +204,15 @@ class PolicyMachine
   #
   # TODO:  might make privilege a class of its own
   def scoped_privileges(user_or_attribute, object_or_attribute, options = {})
-    options = options.dup
-    filters = options.delete(:filters)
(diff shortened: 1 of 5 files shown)


lib/policy_machine.rb
```diff
@@ -95,26 +95,39 @@ def is_privilege?(user_or_attribute, operation, object_or_attribute, options = {
...
+  def is_privilege_with_filters?(user_or_attribute, operation, object_or_attribute, filters: {}, options: {})
+    is_filtered_privilege_ignoring_prohibitions?(user_or_attribute, operation, object_or_attribute, filters: filters, options: options) && (options[:ignore_prohibitions] || !is_privilege_ignoring_prohibitions?(user_or_attribute, PM::Prohibition.on(operation), object_or_attribute, options))
+  end
 
-    unless [PM::Operation, Symbol, String].any? { |allowed_type| operation.is_a?(allowed_type) }
-      raise(ArgumentError, "operation must be an Operation, Symbol, or String.")
-    end
+  ##
+  # Check the privilege with filters without checking for prohibitions.
+  def is_filtered_privilege_ignoring_prohibitions?(user_or_attribute, operation, object_or_attribute, filters: {}, options: {})
```
PoloPro: Can we keep the same method signature? `is_privilege_ignoring_prohibitions_with_filters?`

hstebbins-mdsol: good call, done

hstebbins-mdsol:
Do you mean by further specifying the filter parameter as something like `filters: { user_attributes: { role_uri: role_uri } }` instead of `filters: { role_uri: role_uri }`?

lib/policy_machine.rb
```diff
@@ -95,26 +95,39 @@ def is_privilege?(user_or_attribute, operation, object_or_attribute, options = {
...
+  def is_privilege_with_filters?(user_or_attribute, operation, object_or_attribute, filters: {}, options: {})
+    is_privilege_ignoring_prohibitions_with_filters?(user_or_attribute, operation, object_or_attribute, filters: filters, options: options) && (options[:ignore_prohibitions] || !is_privilege_ignoring_prohibitions?(user_or_attribute, PM::Prohibition.on(operation), object_or_attribute, options))
+  end
 
-    unless [PM::Operation, Symbol, String].any? { |allowed_type| operation.is_a?(allowed_type) }
-      raise(ArgumentError, "operation must be an Operation, Symbol, or String.")
-    end
+  ##
+  # Check the privilege with filters without checking for prohibitions.
+  def is_privilege_ignoring_prohibitions_with_filters?(user_or_attribute, operation, object_or_attribute, filters: {}, options: {})
```
CGDS: So for these PM level public interfaces, I think it would be worthwhile to represent that we're only doing UA filtering.  I think we could either 1) have user attribute specific filtering method names or 2) make the filter args more complicated, so expect keys like `filters: {user_attributes: {key: "val"}}`, which would also allow backwards compatible expansion in future

hstebbins-mdsol: Agreed, I think making the filter args more specific is the best path going forward; fixed

lib/policy_machine.rb
```diff
@@ -95,26 +95,39 @@ def is_privilege?(user_or_attribute, operation, object_or_attribute, options = {
...
   ##
-  # Check the privilege without checking for prohibitions. May be called directly but is also used in is_privilege?
-  #
-  def is_privilege_ignoring_prohibitions?(user_or_attribute, operation, object_or_attribute, options = {})
-    unless user_or_attribute.is_a?(PM::User) || user_or_attribute.is_a?(PM::UserAttribute)
-      raise(ArgumentError, "user_attribute_pe must be a User or UserAttribute.")
-    end
+  # Can we derive a privilege given a set of filters?
+  def is_privilege_with_filters?(user_or_attribute, operation, object_or_attribute, filters: {}, options: {})
+    is_privilege_ignoring_prohibitions_with_filters?(user_or_attribute, operation, object_or_attribute, filters: filters, options: options) && (options[:ignore_prohibitions] || !is_privilege_ignoring_prohibitions?(user_or_attribute, PM::Prohibition.on(operation), object_or_attribute, options))
```
DMcKinnon-mdsol: This line is a little heavy. Mind breaking it up?

lib/policy_machine.rb
```diff
@@ -95,26 +95,39 @@ def is_privilege?(user_or_attribute, operation, object_or_attribute, options = {
...
+  # Check the privilege with filters without checking for prohibitions.
+  def is_privilege_ignoring_prohibitions_with_filters?(user_or_attribute, operation, object_or_attribute, filters: {}, options: {})
+    assert_privilege_parameters!(user_or_attribute, operation, object_or_attribute)
 
-    unless object_or_attribute.is_a?(PM::Object) || object_or_attribute.is_a?(PM::ObjectAttribute)
-      raise(ArgumentError, "object_or_attribute must either be an Object or ObjectAttribute.")
+    if policy_machine_storage_adapter.respond_to?(:is_filtered_privilege?)
+      policy_machine_storage_adapter.is_filtered_privilege?(user_or_attribute, operation, object_or_attribute, filters: filters, options: options)
+    else
+      raise(NotImplementedError, "is_filtered_privilege? not implemented in storage adapter #{policy_machine_storage_adapter.class}")
```
DMcKinnon-mdsol: This is not how this error is supposed to be used: http://chrisstump.online/2016/03/23/stop-abusing-notimplementederror/. Please pick a different error.

lib/policy_machine.rb
```diff
@@ -95,26 +95,39 @@ def is_privilege?(user_or_attribute, operation, object_or_attribute, options = {
...
+  def is_privilege_ignoring_prohibitions?(user_or_attribute, operation, object_or_attribute, options = {})
+    assert_privilege_parameters!(user_or_attribute, operation, object_or_attribute)
 
     if options.empty? && policy_machine_storage_adapter.respond_to?(:is_privilege?)
       privilege = [user_or_attribute, operation, object_or_attribute].map { |obj| obj.respond_to?(:stored_pe) ? obj.stored_pe : obj }
       return policy_machine_storage_adapter.is_privilege?(*privilege)
     end
 
+    if options[:filters] && policy_machine_storage_adapter.respond_to?(:is_privilege_with_filters?)
+      privilege = [user_or_attribute, operation, object_or_attribute].map { |obj| obj.respond_to?(:stored_pe) ? obj.stored_pe : obj }
```
DMcKinnon-mdsol: Why call this list of stored policy elements a privilege?

lib/policy_machine.rb
```diff
@@ -194,17 +207,23 @@ def privileges
   #
   # TODO:  might make privilege a class of its own
   def scoped_privileges(user_or_attribute, object_or_attribute, options = {})
-    privileges_and_prohibitions = if policy_machine_storage_adapter.respond_to?(:scoped_privileges)
-      policy_machine_storage_adapter.scoped_privileges(user_or_attribute.stored_pe, object_or_attribute.stored_pe, options).map do |op|
+    options = options.dup
+    filters = options.delete(:filters)
```
DMcKinnon-mdsol: Using `dup` and `delete` here seems strange to me. Since `filters` is only used once, I'd be find with just calling `options[:filters]` below instead of using `filters`. If you want a separate variable, then I think this would be more standard:
```ruby
filters = options[:filters]
```

spec/policy_machine_storage_adapters/active_record_spec.rb
```diff
@@ -25,6 +25,328 @@
     it_behaves_like 'a policy machine storage adapter'
     let(:policy_machine_storage_adapter) { described_class.new }
 
+    describe 'scoping privileges by user attribute' do
+      let(:candy_pm) { PolicyMachine.new(name: 'Candy Machine', storage_adapter: PolicyMachineStorageAdapter::ActiveRecord)}
```
DMcKinnon-mdsol: Resident unfun person checking in for duty – I don't think fun names make tests any clearer (this is a recent change for me). I think they obfuscate and distract from the specific things that are actually being tested.

I'm sympathetic to the argument that more human names for variables and objects can paint a picture that's easier for some people to follow, but on the whole, I think they engender more confusion than they mitigate.

In any case, this is a personal preference of mine that I just wanted to call out. I don't expect you to change it on this PR.

DMcKinnon-mdsol:
This may have already been discussed and I either missed or forgot the resolution, but can we document here the decision to make prohibitions filter-agnostic. Namely, _for a scoped privilege query, why should it consider prohibitions created outside of the supplied filter?_

spec/policy_machine_storage_adapters/active_record_spec.rb
```diff
@@ -25,6 +25,328 @@
...
+            filters = { color: 'beige' }
+
+            expect(
+              candy_pm.is_privilege_with_filters?(frank, create, brie, filters: filters)
+            ).to be false
+          end
+        end
+
+        context 'when the user does not have access via any user attribute' do
+          it 'returns false' do
```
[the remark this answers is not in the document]
hstebbins-mdsol: Well, the alternative is `it 'returns false when the user does not have access via any user attributes' do`. My reasoning for breaking things out into context blocks here was to keep the usage of context/it blocks idiomatic and concise; if you feel strongly about removing the context blocks here, I'm okay with changing it.

lib/policy_machine.rb
hstebbins-mdsol: Oops, sure thing. I'll keep that in mind going forward.

spec/policy_machine_storage_adapters/active_record_spec.rb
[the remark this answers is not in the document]
DMcKinnon-mdsol: `it 'returns false when the user does not have access via any user attributes' do` seems nice to me haha. But I definitely don't feel strongly about it either way. Feel free to leave it if you'd like.

lib/policy_machine.rb
hstebbins-mdsol: This terminology mirrors the naming when calling `policy_machine_storage_adapter.is_privilege?` directly above. Additionally, there are several places throughout this file where a list of `[user/user_attribute, operation, object/object_attribute]` are referred to as a "privilege" - for example, where `PolicyMachine#privileges` and `PolicyMachine#scoped_privileges` are defined.

spec/policy_machine_storage_adapters/active_record_spec.rb
hstebbins-mdsol: Fair enough, although I think there's a difference between "fun names" and "more human names." Fun names can be distracting, but more human names significantly help me process the graph relationships going on in a lot of these specs.

That being said, I definitely agree that human readable names can easily slip into obfuscation. I'll cool it in future specs.

spec/policy_machine_storage_adapters/active_record_spec.rb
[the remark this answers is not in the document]
hstebbins-mdsol: That's fair, I also don't feel very strongly about it either. Leaving as-is for now

lib/policy_machine_storage_adapters/active_record.rb
```diff
@@ -709,17 +709,31 @@ def is_privilege?(user_or_attribute, operation, object_or_attribute)
...
+    # Returns true if the user has the operation on the object, but only if the privilege
+    # can be derived via a user attribute that passes the filter
+    def is_filtered_privilege?(user_or_attribute, operation, object_or_attribute, filters: {}, options: {})
+      policy_classes_containing_object = policy_classes_for_object_attribute(object_or_attribute)
+      operation_id = operation.try(:unique_identifier) || operation.to_s
+
+      if policy_classes_containing_object.size < 2
+        !accessible_operations(user_or_attribute, object_or_attribute, operation_id, filters: filters).empty?
+      else
+        raise(NotImplementedError, 'is_filtered_privilege? does not support multiple policy classes!')
```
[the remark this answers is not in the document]
hstebbins-mdsol: fixed

DMcKinnon-mdsol:
Merging in 1 hour unless objections @mdsol/team-04.

PR #162, opened 2018-10-16
Performance improvements

hstebbins-mdsol:
This PR is up for general discussion; it's not ready for merge

### DIFF
--- a/lib/policy_machine_storage_adapters/active_record/postgresql.rb
+++ b/lib/policy_machine_storage_adapters/active_record/postgresql.rb
@@ -2,6 +2,42 @@ require 'active_record/hierarchical_query' # via gem activerecord-hierarchical_q
 
 module PolicyMachineStorageAdapter
   class ActiveRecord
+    class PolicyElementAssociation
+      def self.with_accessible_operation(associations, operation)
+        query = <<-SQL
+          operation_set_id IN (
+            WITH RECURSIVE accessible_operations AS (
+              (
+                SELECT
+                  child_id,
+                  parent_id,
+                  parent_id AS operation_set_id
+                FROM assignments
+                WHERE parent_id IN (#{associations.select(:operation_set_id).to_sql})
+              )
+              UNION ALL
+              (
+                SELECT
+                  assignments.child_id,
+                  assignments.parent_id,
+                  accessible_operations.operation_set_id AS operation_set_id
+                FROM assignments
+                INNER JOIN accessible_operations
+                ON accessible_operations.child_id = assignments.parent_id
+              )
+            )
+          SELECT accessible_operations.operation_set_id
+          FROM accessible_operations
+          JOIN policy_elements ops
+            ON ops.id = accessible_operations.child_id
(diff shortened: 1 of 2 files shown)


hstebbins-mdsol:
Sure, running 1000 privilege calls on workshop resulted in a change from ~21ms to ~15ms. I put this up to discuss whether we want to create a story for it, basically, and if I should invest more time in cleaning it up.

DMcKinnon-mdsol:
Yea, seems worth investing the time to clean it up. Might be worth bundling in the same story as the other perf changes I've been poking on the pm. We'll sort that out later.

hstebbins-mdsol:
Sounds good

lib/policy_machine_storage_adapters/active_record/postgresql.rb
```diff
@@ -2,6 +2,42 @@
...
+        query = <<-SQL
+          operation_set_id IN (
+            WITH RECURSIVE accessible_operations AS (
+              (
+                SELECT
+                  child_id,
+                  parent_id,
+                  parent_id AS operation_set_id
+                FROM assignments
+                WHERE parent_id IN (#{associations.select(:operation_set_id).to_sql})
```
fpcyan: is this faster than 
```
  WHERE parent_id IN (:operation_set_ids)
. . .
WHERE ops.unique_identifier = :operation
)
SQL

associations.where(query, operation_set_ids: associations.pluck(:operation_set_id), operation: operation)
```
which i think is more standard rails query

lib/policy_machine_storage_adapters/active_record/postgresql.rb
```diff
@@ -2,6 +2,42 @@
...
+            )
+          SELECT accessible_operations.operation_set_id
+          FROM accessible_operations
+          JOIN policy_elements ops
+            ON ops.id = accessible_operations.child_id
+          WHERE ops.unique_identifier = ?
+          )
+        SQL
+
+        associations.where(query, operation)
```
fpcyan: also maybe it is faster to join directly to the result temp table from the recursive query output instead of us it as a subquery? I'm not sure I remember ever trying that though I've made recursive temp tables 🤔

lib/policy_machine_storage_adapters/active_record/postgresql.rb
hstebbins-mdsol: As is, this implementation compresses that logic into a single query vs two queries (one for the `.pluck` and one when it gets resolved), but I haven't timed the difference.

lib/policy_machine_storage_adapters/active_record/postgresql.rb
hstebbins-mdsol: That's an excellent idea, I would be very surprised if that wasn't more efficient. I'll keep that in mind when I revisit this.

fpcyan: if you do an EXPLAIN_ANALYZE you'll notice that materializing subqueries of arrays of IDs is weirdly slow

PR #164, opened 2019-02-28
[MCC-468035] Move associations_filtered_by_operation to public visibility

PoloPro:
This PR moves the method `associations_filtered_by_operation` to public visibility in the ActiveRecord storage adapter. This method allows policy element associations to be filtered to just those containing the specified operation.

### DIFF
--- a/spec/policy_machine_storage_adapters/active_record_spec.rb
+++ b/spec/policy_machine_storage_adapters/active_record_spec.rb
@@ -1523,6 +1523,70 @@ describe 'ActiveRecord' do
     end
   end
 
+  describe '#associations_filtered_by_operation' do
+    let!(:new_pm) { PolicyMachine.new(name: 'AR PM', storage_adapter: PolicyMachineStorageAdapter::ActiveRecord) }
+
+    let(:operation_1) { new_pm.create_operation('operation_1') }
+    let(:operation_2) { new_pm.create_operation('operation_2') }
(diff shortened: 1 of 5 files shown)


hstebbins-mdsol:
Will merge at noon unless objections @mdsol/team04

DMcKinnon-mdsol:
If they're public now then we should have specs around them.

spec/policy_machine_storage_adapters/active_record_spec.rb
```diff
@@ -1523,6 +1523,70 @@
...
+
+      new_pm.add_assignment(opset_1, opset_2)
+    end
+
+    let(:peas) { PolicyMachineStorageAdapter::ActiveRecord::PolicyElementAssociation.all }
+    let(:pea_1) { peas.find_by(operation_set_id: opset_1.id) }
+    let(:pea_2) { peas.find_by(operation_set_id: opset_2.id) }
+    let(:pea_3) { peas.find_by(operation_set_id: opset_3.id) }
+
+    context 'when provided no policy element associations' do
```
DMcKinnon-mdsol: What's with the "when"s in these context block descriptions? I can't remember seeing you do that before.

DMcKinnon-mdsol: Wes hit me with the http://www.betterspecs.org/

DMcKinnon-mdsol:
Merging in 15 minutes unless objections @mdsol/team-04.

PR #166, opened 2019-04-05
[MCC-468036] Add include_prohibitions filter to #scoped_privileges

PoloPro:
This PR adds an `include_prohibitions` option to the `#scoped_privileges` method in the PolicyMachine. This option returns all prohibitions the user (attribute) has on the specified object (attribute) alongside their derived privileges.

### DIFF
--- a/spec/support/shared_examples_policy_machine.rb
+++ b/spec/support/shared_examples_policy_machine.rb
@@ -1260,23 +1260,24 @@ shared_examples "a policy machine" do
 
       context 'when the user only has prohibitions' do
         let(:all_prohibitions) { policy_machine.create_operation_set('all_prohibitions') }
+        let!(:ua_2) { policy_machine.create_user_attribute('ua_2') }
 
         before do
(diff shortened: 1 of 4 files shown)


spec/support/shared_examples_policy_machine.rb
```diff
@@ -1135,14 +1134,116 @@
...
+        ]
+
+        expect(policy_machine.scoped_privileges(user, one_fish, options)).to match_array(expected_privileges)
+      end
+    end
+
+    context "when the 'include_prohibitions' option is provided" do
+      let(:options) { { include_prohibitions: true } }
+
+      it 'returns all privileges and prohibitions the user has on the specified object' do
```
hstebbins-mdsol: mind adding a spec for behavior when a user is prohibited on an object but doesn't have privilege on it? say:
```
        parent (prohibition)
  child 1 (privilege)   child 2 (no direct assignment)
```

PR #168, opened 2019-04-22
[MCC-488586] Optimize accessible ancestor objects

hstebbins-mdsol:
Hi, this PR adds optimizations to the ActiveRecord Adapter's `accessible_ancestor_objects` method. It moves a lot of the logic from Ruby to raw SQL. In benchmarks with trivial amounts of data, this new method is approximately 3.5x as fast as the old method: 
```ruby
old = lambda { ado_pm.accessible_ancestor_objects(u1, bluff, grandparent_fish, key: :unique_identifier) }
new = lambda { ado_pm.policy_machine_storage_adapter.new_accessible_ancestor_objects(u1, bluff, grandparent_fish, key: :unique_identifier) }
Benchmark.ips do |x|
  x.report("old") { old.call }
  x.report("new") { new.call }
  x.compare!
end
<<-RESULTS
Warming up --------------------------------------
                 old     7.000  i/100ms
                 new    26.000  i/100ms
Calculating -------------------------------------
                 old     74.303  (± 2.7%) i/s -    378.000  in   5.092408s
                 new    264.500  (± 3.0%) i/s -      1.326k in   5.017673s

Comparison:
                 new:      264.5 i/s
                 old:       74.3 i/s - 3.56x  slower
RESULTS
```

@mdsol/team04 - please review

### DIFF
--- a/lib/policy_machine_storage_adapters/active_record/postgresql.rb
+++ b/lib/policy_machine_storage_adapters/active_record/postgresql.rb
@@ -3,7 +3,7 @@ require 'active_record/hierarchical_query' # via gem activerecord-hierarchical_q
 module PolicyMachineStorageAdapter
   class ActiveRecord
     class PolicyElementAssociation
-      def self.all_accessible_objects(associations, root_id: nil, filters: {})
+      def self.all_accessible_objects(associations, root_id:, filters: {})
         pea_ids = associations.pluck(:id)
         return [] if pea_ids.empty?
 
--- a/lib/policy_machine_storage_adapters/active_record/postgresql.rb
+++ b/lib/policy_machine_storage_adapters/active_record/postgresql.rb
@@ -23,7 +23,7 @@ module PolicyMachineStorageAdapter
                 ,a.parent_id
               FROM assignments a
               JOIN ancestors anc
-      ON anc.parent_id = a.child_id
+                ON anc.parent_id = a.child_id
             )
             ),
 
--- a/lib/policy_machine_storage_adapters/active_record/postgresql.rb
+++ b/lib/policy_machine_storage_adapters/active_record/postgresql.rb
@@ -10,51 +10,79 @@ module PolicyMachineStorageAdapter
         query = <<-SQL
           id IN (
             WITH RECURSIVE ancestor_scope AS (
-            (
-              SELECT
-                child_id AS parent_id,
-                child_id AS child_id
-              FROM assignments
-              WHERE child_id = ?
-            )
+            SELECT
+                child_id,
+                child_id AS parent_id
+            FROM assignments
+            WHERE child_id = ?
+
             UNION ALL
-            (
-              SELECT
+
+            SELECT
+                assignments.child_id,
+                assignments.parent_id
+            FROM assignments
+            JOIN ancestor_scope
+                ON assignments.child_id = ancestor_scope.parent_id
+            ),
+
+            leaf_ancestors AS (
+            SELECT DISTINCT
+                ancestor_scope.parent_id AS parent_id
+            FROM ancestor_scope
+            LEFT OUTER JOIN assignments
+                ON ancestor_scope.parent_id = assignments.child_id
+            WHERE assignments.child_id IS NULL
+            ),
+
+            leaf_descendants AS (
+            SELECT
+                parent_id,
+                parent_id AS child_id
+            FROM leaf_ancestors
+
+            UNION ALL
+
+            SELECT
                 assignments.parent_id,
                 assignments.child_id
-              FROM assignments
-              JOIN ancestor_scope
-                ON ancestor_scope.parent_id = assignments.child_id
-            )
+            FROM assignments
+            JOIN leaf_descendants
+                ON leaf_descendants.child_id = assignments.parent_id
+            ),
+
+            accessible_leaf_descendants AS (
+            SELECT DISTINCT
+                leaf_descendants.child_id
+            FROM leaf_descendants
+            JOIN policy_element_associations peas
+                ON peas.object_attribute_id = leaf_descendants.child_id
+            WHERE 1=1
+                AND peas.id IN (#{pea_ids.join(',')})
             ),
 
             accessible_ancestors AS (
-            (
-              SELECT
-                ARRAY[peas.object_attribute_id] AS path
-                ,peas.object_attribute_id AS parent_id
-              FROM policy_element_associations peas
-              WHERE peas.id IN (#{pea_ids.join(',')})
-            )
+            SELECT
+                child_id,
+                child_id AS parent_id
+            FROM accessible_leaf_descendants
+
             UNION ALL
-            (
-              SELECT
-                path || assignments.parent_id AS path
-                ,assignments.parent_id AS parent_id
-              FROM assignments
-              JOIN accessible_ancestors
-                ON accessible_ancestors.parent_id = assignments.child_id
-            )
+
+            SELECT
+                a.child_id,
+                a.parent_id
+            FROM assignments a
+            JOIN accessible_ancestors d
+                ON a.child_id = d.parent_id
             )
 
-            SELECT DISTINCT
-              ancestor_scope.parent_id
-            FROM ancestor_scope
-            JOIN accessible_ancestors
-              ON ancestor_scope.parent_id = accessible_ancestors.parent_id
-                AND ARRAY[ancestor_scope.parent_id] <@ accessible_ancestors.path
-
-            )
+            SELECT
+                accessible_ancestors.parent_id
+            FROM accessible_ancestors
+            JOIN ancestor_scope
+                ON ancestor_scope.parent_id = accessible_ancestors.parent_id
+          )
         SQL
 
         PolicyElement.where(query, root_id).where(filters)
--- a/lib/policy_machine_storage_adapters/active_record/postgresql.rb
+++ b/lib/policy_machine_storage_adapters/active_record/postgresql.rb
@@ -6,7 +6,7 @@ module PolicyMachineStorageAdapter
       def self.all_accessible_objects(associations, root_id: nil)
         query = <<-SQL
           id IN (
-            WITH RECURSIVE ancestors AS (
+            WITH RECURSIVE accessible_ancestors AS (
             (
               SELECT
                 objects.id AS parent_id,
@@ -19,11 +19,11 @@ module PolicyMachineStorageAdapter
             UNION ALL
             (
               SELECT
-                a.parent_id,
-                a.parent_id
-              FROM assignments a
-              JOIN ancestors anc
-                ON anc.parent_id = a.child_id
+                assignments.parent_id,
+                assignments.parent_id
+              FROM assignments
+              JOIN accessible_ancestors
+                ON accessible_ancestors.parent_id = assignments.child_id
             )
             ),
 
@@ -38,17 +38,17 @@ module PolicyMachineStorageAdapter
             UNION ALL
             (
               SELECT
-                a.parent_id,
-                a.child_id
-              FROM assignments a
-              JOIN ancestor_scope anc
-                ON anc.parent_id = a.child_id
+                assignments.parent_id,
+                assignments.child_id
+              FROM assignments
+              JOIN ancestor_scope
+                ON ancestor_scope.parent_id = assignments.child_id
             )
             )
 
             SELECT DISTINCT
               parent_id
-            FROM ancestors
+            FROM accessible_ancestors
             WHERE parent_id IN (SELECT parent_id FROM ancestor_scope)
           )
         SQL
--- a/lib/policy_machine_storage_adapters/active_record/postgresql.rb
+++ b/lib/policy_machine_storage_adapters/active_record/postgresql.rb
@@ -3,16 +3,19 @@ require 'active_record/hierarchical_query' # via gem activerecord-hierarchical_q
 module PolicyMachineStorageAdapter
   class ActiveRecord
     class PolicyElementAssociation
-      def self.all_accessible_objects(associations, root_id: nil)
+      def self.all_accessible_objects(associations, root_id: nil, filters: {})
+        pea_ids = associations.pluck(:id)
+        return [] if pea_ids.empty?
+
         query = <<-SQL
           id IN (
             WITH RECURSIVE ancestor_scope AS (
             (
               SELECT
-                id AS parent_id,
-                id AS child_id
-              FROM policy_elements
-              WHERE id = ?
+                child_id AS parent_id,
+                child_id AS child_id
+              FROM assignments
+              WHERE child_id = ?
             )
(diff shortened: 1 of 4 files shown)


lib/policy_machine_storage_adapters/active_record/postgresql.rb
```diff
@@ -2,6 +2,95 @@
...
   class ActiveRecord
+    class PolicyElementAssociation
+      def self.all_accessible_objects(associations, root_id: nil)
+        query = <<-SQL
+          id IN (
+            WITH RECURSIVE ancestors AS (
+            (
+              SELECT
+                objects.id AS parent_id
+                ,objects.id AS child_id
```
fpcyan: why are the commas at the start of lines?

lib/policy_machine_storage_adapters/active_record/postgresql.rb
```diff
@@ -2,6 +2,95 @@
...
+              JOIN policy_elements objects
+                ON objects.id = peas.object_attribute_id
+              WHERE peas.id IN (#{associations.select(:id).to_sql})
+            )
+            UNION ALL
+            (
+              SELECT
+                a.parent_id
+                ,a.parent_id
+              FROM assignments a
```
fpcyan: your aliasing is kind of inconsistent

lib/policy_machine_storage_adapters/active_record/postgresql.rb
```diff
@@ -2,6 +2,95 @@
...
+        PolicyElement.where(query, root_id)
+      end
+
+      def self.with_accessible_operation(associations, operation)
+        query = <<-SQL
+          operation_set_id IN (
+           WITH RECURSIVE accessible_operations AS (
+              (
+               SELECT
+                  child_id,
```
fpcyan: now the commas are in the right place

lib/policy_machine_storage_adapters/active_record/postgresql.rb
```diff
@@ -2,6 +2,95 @@
 
 module PolicyMachineStorageAdapter
   class ActiveRecord
+    class PolicyElementAssociation
+      def self.all_accessible_objects(associations, root_id: nil)
```
fpcyan: I'm a little skeptical that a sql method that does two recursive functions is going to be faster on big data. I'd be curious about performance on that

lib/policy_machine_storage_adapters/active_record/postgresql.rb
hstebbins-mdsol: i do this when writing queries because it's easier for me to tell when i'm missing a comma when adding/removing columns. it makes a big difference for my sql development speed, weirdly enough, but it's just a development quirk that i forgot to fix before i submitted the PR. sadly, no one else likes it but me 😢

lib/policy_machine_storage_adapters/active_record/postgresql.rb
hstebbins-mdsol: very fair, i'll do some sandbox performance testing with highly assigned users and update when i get more data

lib/policy_machine_storage_adapters/active_record/postgresql.rb
hstebbins-mdsol: fair enough, i'll update. lemme know if you think it's still unclear/inconsistent

hstebbins-mdsol:
dnm until i get performance results with more highly connected data

DMcKinnon-mdsol:
> dnm until i get performance results with more highly connected data

Please also open a PR against dalton using this branch so we can see those specs pass too before we merge.

lib/policy_machine_storage_adapters/active_record/postgresql.rb
fpcyan: okay that's actually really very clever. i like it now

hstebbins-mdsol:
Dalton PR: https://github.com/mdsol/dalton/pull/3016

hstebbins-mdsol:
@fpcyan was right! 😅 the existing implementation was quite slow against big data.

I've further modified the method to improve performance against large sets of data.

I've updated the OP, but now it's roughly 3.5x fast against trivial data. That being said, it's now faster against large amounts of data. These numbers are from a heavily assigned user (King Jim), getting scoped `read` privileges across 10 randomly selected CDSs he's assigned within. Note that this includes both direct assignments to CDSs and assignments to objects within a CDS.

```ruby
{"<cds uri>"=>{"ancestor_count"=>25, "result"=>{"old"=>1.53, "new"=>2.35}},
 "<cds uri>"=>{"ancestor_count"=>81, "result"=>{"old"=>1.82, "new"=>2.48}},
 "<cds uri>"=>{"ancestor_count"=>2092, "result"=>{"old"=>1.55, "new"=>1.67}},
 "<cds uri>"=>{"ancestor_count"=>63, "result"=>{"old"=>1.93, "new"=>2.52}},
 "<cds uri>"=>{"ancestor_count"=>1445},
 "<cds uri>"=>{"ancestor_count"=>76, "result"=>{"old"=>1.89, "new"=>2.45}},
 "<cds uri>"=>{"ancestor_count"=>311, "result"=>{"old"=>1.88, "new"=>2.48}},
 "<cds uri>"=>{"ancestor_count"=>1473, "result"=>{"old"=>1.34, "new"=>1.76}},
 "<cds uri>"=>{"ancestor_count"=>127, "result"=>{"old"=>1.84, "new"=>2.45}}}
```
Sorry for the raw data dump. Also, note that `result` values are for Iterations Per Second, _not_ elapsed time. So, a higher number == better performance.

The performance across the board is between 1.08x to 1.53x better with this new method, but 7/10 of these are around 1.30x better.

Note that one CDS is missing values: this is because attempting to execute the old method against that CDS results in a timeout (I'm testing against the read replica, which is much more sensitive to timeouts). That being said, the new method returns in 523ms.

Sorry for the delay in getting back to this, but I wanted to make sure I was getting good performance data. I now feel pretty confident about these performance improvements across the board in terms of data size.

lib/policy_machine_storage_adapters/active_record/postgresql.rb
```diff
@@ -2,6 +2,127 @@
 
 module PolicyMachineStorageAdapter
   class ActiveRecord
+    class PolicyElementAssociation
+      def self.all_accessible_objects(associations, root_id:, filters: {})
```
DMcKinnon-mdsol: How about just naming this `accessible_objects`?

lib/policy_machine_storage_adapters/active_record/postgresql.rb
```diff
@@ -2,6 +2,127 @@
...
+        pea_ids = associations.pluck(:id)
+        return [] if pea_ids.empty?
```
DMcKinnon-mdsol: As is, this can return different _types_ of objects – either an active record association of policy elements or arrays. If `.pluck` returns an active record association, then can we return that here to be consistent? Like...
```ruby
return pea_ids if pea_ids.empty?
```

lib/policy_machine_storage_adapters/active_record/postgresql.rb
```diff
@@ -2,6 +2,127 @@
...
+
+        query = <<-SQL
+          id IN (
+            WITH RECURSIVE ancestor_scope AS (
+            SELECT
```
DMcKinnon-mdsol: Preference to indent within the parenthesis like you did with `WITH RECURSIVE ancestor_scope AS (`.

DMcKinnon-mdsol: Same comment below.

lib/policy_machine_storage_adapters/active_record/postgresql.rb
```diff
@@ -2,6 +2,127 @@
 
 module PolicyMachineStorageAdapter
   class ActiveRecord
+    class PolicyElementAssociation
+      def self.all_accessible_objects(associations, root_id:, filters: {})
+        pea_ids = associations.pluck(:id)
+        return [] if pea_ids.empty?
+
+        query = <<-SQL
```
DMcKinnon-mdsol: Comment please.

lib/policy_machine_storage_adapters/active_record/postgresql.rb
```diff
@@ -2,6 +2,127 @@
...
+            FROM assignments
+            JOIN ancestor_scope
+                ON assignments.child_id = ancestor_scope.parent_id
+            ),
+
+            leaf_ancestors AS (
+            SELECT DISTINCT
+                ancestor_scope.parent_id AS parent_id
+            FROM ancestor_scope
+            LEFT OUTER JOIN assignments
```
DMcKinnon-mdsol: Does this need to be a left outer join? I know we've seen performance issues around left outers that we can avoid before.

lib/policy_machine_storage_adapters/active_record/postgresql.rb
```diff
@@ -2,6 +2,127 @@
...
+            SELECT
+                assignments.child_id,
+                assignments.parent_id
+            FROM assignments
+            JOIN ancestor_scope
+                ON assignments.child_id = ancestor_scope.parent_id
+            ),
+
+            leaf_ancestors AS (
+            SELECT DISTINCT
```
DMcKinnon-mdsol: Do we need these distincts? In particular, I'm curious if postgres handles this duplication for us while recursing and leaves us with a performance boost if we remove the distinct because we wouldn't have to do this (potentially) redundant unique on every iteration of the recursion. Same comment below.

lib/policy_machine_storage_adapters/active_record/postgresql.rb
```diff
@@ -2,6 +2,127 @@
...
+                ON leaf_descendants.child_id = assignments.parent_id
+            ),
+
+            accessible_leaf_descendants AS (
+            SELECT DISTINCT
+                leaf_descendants.child_id
+            FROM leaf_descendants
+            JOIN policy_element_associations peas
+                ON peas.object_attribute_id = leaf_descendants.child_id
+            WHERE 1=1
```
DMcKinnon-mdsol: Comment please.

lib/policy_machine_storage_adapters/active_record/postgresql.rb
```diff
@@ -2,6 +2,127 @@
...
+                AND peas.id IN (#{pea_ids.join(',')})
```
DMcKinnon-mdsol: You used `?` for interpolation above and below. Any reason not to be consistent with the interpolation method?

lib/policy_machine_storage_adapters/active_record/postgresql.rb
```diff
@@ -2,6 +2,127 @@
...
+                  assignments.parent_id,
+                 accessible_operations.operation_set_id AS operation_set_id
+                FROM assignments
+                INNER JOIN accessible_operations
+                ON accessible_operations.child_id = assignments.parent_id
+             )
+            )
+          SELECT accessible_operations.operation_set_id
+          FROM accessible_operations
+         JOIN policy_elements ops
```
DMcKinnon-mdsol: There's weird formatting around this line.

lib/policy_machine_storage_adapters/active_record/postgresql.rb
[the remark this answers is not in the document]
hstebbins-mdsol: I dislike both names because this requires a root scope. How about `scoped_accessible_objects`?

lib/policy_machine_storage_adapters/active_record/postgresql.rb
hstebbins-mdsol: how about the ole `PolicyElement.none`

lib/policy_machine_storage_adapters/active_record/postgresql.rb
hstebbins-mdsol: What do you suggest as an alternative? I can't really think of a different way to do it.

lib/policy_machine_storage_adapters/active_record/postgresql.rb
```diff
@@ -2,6 +2,127 @@
 
 module PolicyMachineStorageAdapter
   class ActiveRecord
+    class PolicyElementAssociation
+      def self.all_accessible_objects(associations, root_id:, filters: {})
+        pea_ids = associations.pluck(:id)
```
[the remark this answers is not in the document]
hstebbins-mdsol: No, I don't think so. I can't think of a usecase involving a PEA relation with multiple IDs. Additionally, it's used in a `id IN (...)` clause, which doesn't differentiate between unique sets and non-unique sets.

lib/policy_machine_storage_adapters/active_record/postgresql.rb
hstebbins-mdsol: fine, fine, i'll remove more of my query quirks

lib/policy_machine_storage_adapters/active_record/postgresql.rb
DMcKinnon-mdsol: Talked offline. I misread this. It's fine as is 👍 .

lib/policy_machine_storage_adapters/active_record/postgresql.rb
DMcKinnon-mdsol: If this come with a performance improvement because of some weird psql thing then we should keep it (and add a comment). I just can't remember if that's the case.

lib/policy_machine_storage_adapters/active_record/postgresql.rb
hstebbins-mdsol: These distincts are intentional for performance reasons.

The key here is the use of `UNION ALL` in the recursive CTEs. Using `UNION ALL` does _not_ enforce uniqueness, while using `UNION` does. Using `UNION` has worse performance and using `UNION ALL` leaves duplicates. By selecting distinct records _outside_ the query, we get the performance benefits of not uniqueness checking on each recursion while still getting a unique set. Having a unique set is useful because we won't have to duplicate work in recursive queries using these results.

lib/policy_machine_storage_adapters/active_record/postgresql.rb
hstebbins-mdsol: nope, just a quirk in how i write SQL.

lib/policy_machine_storage_adapters/active_record/postgresql.rb
hstebbins-mdsol: nope, will change

lib/policy_machine_storage_adapters/active_record/postgresql.rb
hstebbins-mdsol: done

hstebbins-mdsol: let me know if it's not sufficient!

PR #171, opened 2019-07-25
[MCC-521095] Allow preloading of policy element associations in .accessible_ancestor_objects

PoloPro:
This PR updates `.accessible_ancestor_objects` to support an optional argument `associations_with_operation`. This bypasses the loading of PolicyElementAssociations for the user with the specified operation, if the relation has already been loaded.

Also updates the `options` argument to a keyword argument to reflect the optional `associations_with_operation`.

### DIFF
--- a/lib/policy_machine_storage_adapters/active_record.rb
+++ b/lib/policy_machine_storage_adapters/active_record.rb
@@ -774,17 +774,19 @@ module PolicyMachineStorageAdapter
 
     # Version of accessible_objects which only returns objects that are
     # ancestors of a specified root object or the object itself
-    def accessible_ancestor_objects(user_or_attribute, operation, root_object, options = {})
+    def accessible_ancestor_objects(user_or_attribute, operation, root_object, associations_with_operation = nil, options: {})
       # If the root_object is a generic PM::Object, convert it the appropriate storage adapter Object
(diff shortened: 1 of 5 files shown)


lib/policy_machine_storage_adapters/active_record.rb
```diff
@@ -772,19 +772,22 @@ def accessible_objects(user_or_attribute, operation, options = {})
       end
     end
 
-    # Version of accessible_objects which only returns objects that are
-    # ancestors of a specified root object or the object itself
-    def accessible_ancestor_objects(user_or_attribute, operation, root_object, options = {})
+    # Version of accessible_objects which only returns objects that are ancestors of a specified
+    # root object or the object itself. A set of policy element associations with the specified
+    # operation may be optionally provided.
+    def accessible_ancestor_objects(user_or_attribute, operation, root_object, associations_with_operation = nil, options: {})
```
hstebbins-mdsol: Would it be possible to make `associations_with_operation` a key in the options hash?

I think this would be ideal because we also wouldn't technically break backwards compatibility since we wouldn't have to keyword the options hash. 

Plus, it just feels a little more straightforward to have `associations_with_operation` identifier by a keyword when used.

PR #174, opened 2020-04-20
[MCC-530336] Update Policy Machine to Rails 6

edkim-mdsol:
Rails 6 has new features that can be viewed here: 
https://prograils.com/posts/new-features-rails-6-multiple-databases-parallel-tests-action-mailbox-etc
The changelog for Rails 6 can be viewed here:
https://github.com/rails/rails/blob/6-0-stable/railties/CHANGELOG.md

This change was not compatible with older ruby versions and required Ruby 2.5.0+.


test/tasks/setup.rake
```diff
@@ -5,7 +5,7 @@ namespace :pm do
       Dir.chdir('./test') do
         db_type = args[:db_type] || 'postgresql'
         `rm -rf testapp`
-        `bundle exec rails new testapp -f -d #{db_type} -m ./template.rb  --skip-keeps --skip-spring  --skip-git`
+        `bundle exec rails new testapp -f -d #{db_type} -m ./template.rb  --skip-keeps --skip-spring  --skip-git --skip-javascript`
```
apejovich, line 8: why are we skipping javascript here?

edkim-mdsol: Rails 6 by default uses Webpacker which was causing some errors so we added that to disable it.

DMcKinnon-mdsol:
Has this been tested on sandbox?

PR #176, opened 2020-06-25
[MCC-634582] Check prohibitions before checking if a priv exists

edkim-mdsol:
Background
- When performing a privilege exists lookup, Dalton (via the policy machine) checks if an operator is prohibited on the requested operable after it has checked if the user has access the to the given operable via the requested permission.

Changes
- Check prohibitions first before performing privilege derivation.

Tests
- Sample size of 1000
- Setting `ignore_prohibitions` to `true` results in an average response time of 1.42751863352 seconds
- Setting `ignore_prohibitions`to `false` results in an average response time of 2.19180535417 seconds 

Setting `ignore_prohibitions` to `true` is ~53.54% faster.

### DIFF
--- a/lib/policy_machine.rb
+++ b/lib/policy_machine.rb
@@ -90,16 +90,18 @@ class PolicyMachine
   # TODO: add option to ignore policy classes to allow consumer to speed up this method.
   # TODO: Parallelize the two component checks
   def is_privilege?(user_or_attribute, operation, object_or_attribute, options = {})
+    options_copy = options.deep_dup
     (options[:ignore_prohibitions] || !is_privilege_ignoring_prohibitions?(user_or_attribute, PM::Prohibition.on(operation), object_or_attribute, options)) &&
-      is_privilege_ignoring_prohibitions?(user_or_attribute, operation, object_or_attribute, options)
+      is_privilege_ignoring_prohibitions?(user_or_attribute, operation, object_or_attribute, options_copy)
   end
 
   ##
   # Can we derive a privilege given a set of filters?
   def is_privilege_with_filters?(user_or_attribute, operation, object_or_attribute, filters: {}, options: {})
     # Check that the privilege can be derived given the set of filters, but do not filter the check for prohibitions
+    options_copy = options.deep_dup
     (options[:ignore_prohibitions] || !is_privilege_ignoring_prohibitions?(user_or_attribute, PM::Prohibition.on(operation), object_or_attribute, options)) &&
-      is_privilege_ignoring_prohibitions_with_filters?(user_or_attribute, operation, object_or_attribute, filters: filters, options: options)
+      is_privilege_ignoring_prohibitions_with_filters?(user_or_attribute, operation, object_or_attribute, filters: filters, options: options_copy)
   end
 
   ##
--- a/lib/policy_machine.rb
+++ b/lib/policy_machine.rb
@@ -90,9 +90,7 @@ class PolicyMachine
   # TODO: add option to ignore policy classes to allow consumer to speed up this method.
   # TODO: Parallelize the two component checks
   def is_privilege?(user_or_attribute, operation, object_or_attribute, options = {})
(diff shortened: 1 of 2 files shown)


lib/policy_machine.rb
```diff
@@ -90,16 +90,18 @@ def add_association(user_attribute_pe, operation_set, object_attribute_pe)
   # TODO: add option to ignore policy classes to allow consumer to speed up this method.
   # TODO: Parallelize the two component checks
   def is_privilege?(user_or_attribute, operation, object_or_attribute, options = {})
-    is_privilege_ignoring_prohibitions?(user_or_attribute, operation, object_or_attribute, options) &&
-      (options[:ignore_prohibitions] || !is_privilege_ignoring_prohibitions?(user_or_attribute, PM::Prohibition.on(operation), object_or_attribute, options))
+    options_copy = options.deep_dup
```
DMcKinnon-mdsol, line 93: WHy is this necessary?

edkim-mdsol: For some reason, the options hash values were wiping themselves after running `!is_privilege_ignoring_prohibitions?(user_or_attribute, PM::Prohibition.on(operation), object_or_attribute, options)` This was a quickfix to see if any other specs were broken. Should I investigate the cause or this is a viable solution?

lib/policy_machine.rb
```diff
@@ -90,16 +90,19 @@ def add_association(user_attribute_pe, operation_set, object_attribute_pe)
   # TODO: add option to ignore policy classes to allow consumer to speed up this method.
   # TODO: Parallelize the two component checks
   def is_privilege?(user_or_attribute, operation, object_or_attribute, options = {})
-    is_privilege_ignoring_prohibitions?(user_or_attribute, operation, object_or_attribute, options) &&
-      (options[:ignore_prohibitions] || !is_privilege_ignoring_prohibitions?(user_or_attribute, PM::Prohibition.on(operation), object_or_attribute, options))
+    # Fixes situation where options hash's values are deleting after first is_privilege_ignoring_prohibitions? call.
+    options_copy = options.deep_dup
+    (options[:ignore_prohibitions] || !is_privilege_ignoring_prohibitions?(user_or_attribute, PM::Prohibition.on(operation), object_or_attribute, options)) &&
+      is_privilege_ignoring_prohibitions?(user_or_attribute, operation, object_or_attribute, options_copy)
```
DMcKinnon-mdsol, line 96: Preference for passing the copy into the prohibitions method if that's what's being mutated.
```suggestion
    (options[:ignore_prohibitions] || !is_privilege_ignoring_prohibitions?(user_or_attribute, PM::Prohibition.on(operation), object_or_attribute, options_copy)) &&
      is_privilege_ignoring_prohibitions?(user_or_attribute, operation, object_or_attribute, options)
```

lib/policy_machine.rb
DMcKinnon-mdsol: YIkes. Okay, you don't need to investigate this, but please add a comment.

I would _prefer_ that you investigated and fixed the consuming method not to mutate its input, though :)

edkim-mdsol: I'll take a look!

lib/policy_machine.rb
```diff
@@ -137,7 +137,7 @@ def is_privilege_ignoring_prohibitions?(user_or_attribute, operation, object_or_
       raise(ArgumentError, "options[:associations] cannot be empty") if associations.empty?
       raise(ArgumentError, "expected each element of options[:associations] to be a PM::Association") unless associations.all?{|a| a.is_a?(PM::Association)}
 
-      associations.keep_if do |association|
+      associations = associations.select do |association|
         association.operation_set.connected?(operation)
       end
       return false if associations.empty?
```
DMcKinnon-mdsol, line 143:
```suggestion
      return false if associations.none? do |association|
        association.operation_set.connected?(operation)
      end
```

PR #178, opened 2020-08-18
[MCC-651906] Add pluck method to Policy Machine and consume in at least one place in Dalton

okashansky-mdsol:
[MCC-651906] Add pluck method to Policy Machine and consume in at least one place in Dalton

Background
We need one of these for performance reasons, namely so that we can only select a single column or subset of columns from the database (instead of all the columns).

Changes Summary
The changes implemented and tested for  **PolicyMachineStorageAdapter::ActiveRecord** storage adapter. 
This is the PM-related part of the changes. Dalton-related changes are in Dalton repo.

* **pluck** method added to **PolicyMachineStorageAdapter::ActiveRecord**
* **pluck** method added to **PolicyMachine**
* specs added to **active_record_spec**

Scenarios
We can get an array of entity fields instead of ActiveRecord-descendant entity with code like:
```
<policy-machine-instance>.pluck(type: :type, fields: [<fields>], options: {<options> } )
```
or in Dalton (please use a branch in which implemented   ActivePolicyElement::Base.pluck)
```
<ActivePolicyElement::Base-descendant-class>.pluck(fields: [<fields>], options: {<options> } )
```


.gitignore
```diff
@@ -1,3 +1,5 @@
 test/testapp/
 Gemfile.lock
 coverage/
+.idea/
+.vscode/
```
DMcKinnon-mdsol, line 5:
```suggestion
.vscode/

```

.idea/.gitignore
```diff
@@ -0,0 +1,10 @@
+# # Default ignored files
+# /shelf/
+# /workspace.xml
+# # Datasource local storage ignored files
+# /dataSources/
+# /dataSources.local.xml
+# # Editor-based HTTP Client requests
+# /httpRequests/
+.vscode/
+.idea/
```
DMcKinnon-mdsol, line 10: Is this file necessary?

lib/policy_machine.rb
```diff
@@ -244,6 +244,15 @@ def batch_find(type:, query: {}, config: {}, &blk)
     end
   end
 
+  def pluck(type:, fields:, options:)
+    if policy_machine_storage_adapter.respond_to?(:pluck)
+      policy_machine_storage_adapter.send(:pluck, {type: type, fields: fields, options: options})
+    else
+      Warn.once("WARNING: 'pluck' is not implemented for storage adapter #{policy_machine_storage_adapter.class}")
+      nil
```
DMcKinnon-mdsol, line 252: I think we should actually hard raise here.

lib/policy_machine.rb
```diff
@@ -244,6 +244,15 @@ def batch_find(type:, query: {}, config: {}, &blk)
     end
   end
 
+  def pluck(type:, fields:, options:)
+    if policy_machine_storage_adapter.respond_to?(:pluck)
+      policy_machine_storage_adapter.send(:pluck, {type: type, fields: fields, options: options})
```
DMcKinnon-mdsol, line 249:
```suggestion
      policy_machine_storage_adapter.send(:pluck, { type: type, fields: fields, options: options })
```

lib/policy_machine_storage_adapters/active_record.rb
```diff
@@ -489,6 +489,12 @@ def operations
       end
     end # End of POLICY_ELEMENT_TYPES iteration
 
+    def pluck(type:, fields: [], options: {})
```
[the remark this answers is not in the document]
DMcKinnon-mdsol: Oh, I see what you're doing now! Nevermind!

spec/policy_machine_storage_adapters/active_record_spec.rb
```diff
@@ -92,7 +92,7 @@
             filters = { user_attributes: { color: 'purple' } }
 
             expect(
-              priv_pm.is_privilege_with_filters?(user_1, create, object_5, filters: filters)
+                priv_pm.is_privilege_with_filters?(user_1, create, object_5, filters: filters)
```
DMcKinnon-mdsol, line 95: Why did you make all these spacing changes? 2 spaces is the correct default.

spec/policy_machine_storage_adapters/active_record_spec.rb
```diff
@@ -1697,6 +1697,17 @@
           end
         end
       end
+
+      describe 'pluck' do
+        (PolicyMachine::POLICY_ELEMENT_TYPES).each do |type|
+          it "can get data for #{type}" do
```
DMcKinnon-mdsol, line 1703:
```suggestion
          it "plucks the correct data for #{type}" do
```

spec/policy_machine_storage_adapters/active_record_spec.rb
```diff
@@ -1697,6 +1697,17 @@
...
+            id = "#{type}-pluck-test"
+            policy_machine.send("create_#{type}", id)
+            data = policy_machine.send(:pluck, type: type, fields: [:unique_identifier], options: { unique_identifier: id })
+            expect(data[0]).to eq(id)
```
DMcKinnon-mdsol, line 1707: I would actually test the entire data structure like
```suggestion
            expect(data).to eq([id])
```

spec/policy_machine_storage_adapters/active_record_spec.rb
okashansky-mdsol: I used to think that auto-formatting is basic thing of any development environment in XXI century. I was wrong...

spec/policy_machine_storage_adapters/active_record_spec.rb
okashansky-mdsol: updated

.idea/.gitignore
okashansky-mdsol: > Looking good! Mind also opening a pull request in Dalton that uses this branch and the pluck functionality?

I guess first we need to merge it to master? and create a tag. After that change Dalton's Gemfile probably.

lib/policy_machine.rb
okashansky-mdsol: Now it throws the error.

lib/policy_machine.rb
```diff
@@ -244,6 +244,14 @@ def batch_find(type:, query: {}, config: {}, &blk)
     end
   end
 
+  def pluck(type:, fields:, options:)
+    if policy_machine_storage_adapter.respond_to?(:pluck)
+      policy_machine_storage_adapter.send(:pluck, { type: type, fields: fields, options: options })
+    else
+      throw MethodNotSupportedError.new("Method 'pluck' is not implemented for storage adapter #{policy_machine_storage_adapter.class}")
```
DMcKinnon-mdsol, line 251: `throw` is extremely rare in Ruby. I've only actually seen it used in a handful of codebases. When it's used, iirc it should really only be used with catch. Otherwise we should probably just use `raise`. I think that's the case here. Let's just use `raise`.

DMcKinnon-mdsol:
Please update the changelog file with a description of this change.

spec/policy_machine_storage_adapters/active_record_spec.rb
```diff
@@ -1697,6 +1697,17 @@
           end
         end
       end
+
+      describe 'pluck' do
+        (PolicyMachine::POLICY_ELEMENT_TYPES).each do |type|
```
DMcKinnon-mdsol, line 1702: You don't need parenthesis here.

lib/policy_machine.rb
okashansky-mdsol: fixed

spec/policy_machine_storage_adapters/active_record_spec.rb
okashansky-mdsol: fixed

PR #179, opened 2020-08-20
Revert "[MCC-634582] Check prohibitions before checking if a priv exists"

DMcKinnon-mdsol:
Reverts mdsol/the_policy_machine#176


DMcKinnon-mdsol:
@mdsol/team04 please review.

DMcKinnon-mdsol:
Yep. We don't delete tags that we've already put out into the wild. That could break unexpected things.

PR #182, opened 2020-09-17
[MCC-631454] Directly Accessible Objects

ejinotti-mdsol:
@mdsol/c2d2

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,7 @@
 # Changelog
 
 ## 3.1.0
-* Add `direct_only` option to `PolicyMachineStorageAdapter::ActiveRecord::accessible_objects` to fetch only directly assigned objects.
+* Added `direct_only` option to `PolicyMachineStorageAdapter::ActiveRecord::accessible_objects` to fetch only directly assigned objects.
 
 ## 3.0.2
 * Added `pluck` method to PolicyMachine. Implemented for PolicyMachineStorageAdapter::ActiveRecord only.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,8 @@
 # Changelog
 
+## 3.1.0
+* Add `direct_only` option to `PolicyMachineStorageAdapter::ActiveRecord::accessible_objects` to fetch only directly assigned objects.
+
 ## 3.0.2
 * Added `pluck` method to PolicyMachine. Implemented for PolicyMachineStorageAdapter::ActiveRecord only.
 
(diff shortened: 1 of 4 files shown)


CHANGELOG.md
```diff
@@ -1,5 +1,8 @@
 # Changelog
 
+## 3.1.0
+* Added `direct_only` option to `PolicyMachineStorageAdapter::ActiveRecord::accessible_objects` to fetch only directly assigned objects.
+
```
ejinotti-mdsol, line 5: FWIW, when I bumped the minor version I wasn't sure what to bump.. looks like for the pluck method they just bumped tiny so perhaps this should be 3.0.3 rather than 3.1.0?  I don't really have an opinion but wanted to point out for discussion.

CGDS: Think that's fine for our purposes

lib/policy_machine/version.rb
```diff
@@ -1,3 +1,3 @@
 class PolicyMachine
-  VERSION = "2.1.3"
+  VERSION = "3.1.0"
```
CGDS, line 2: hadn't kept that up to date in a while i see lol

lib/policy_machine_storage_adapters/active_record.rb
```diff
@@ -846,17 +846,20 @@ def associations_for_user_or_attribute(user_or_attribute, options)
...
+      # Indirect scope: the set of objects which the operator can access via ancestral hierarchy
+      unless options[:direct_only]
+        scopes << Assignment.ancestors_of(permitting_oas).where(type: class_for_type('object').name)
       end
 
-      direct_scope | indirect_scope
+      inclusion = options[:includes]
+      scopes.map! { |s| build_inclusion_scope(s, options[:key], inclusion) } if inclusion
+
+      scopes.reduce(&:|).to_a
```
CGDS, line 862: Think for a user of this interface external to the PM, a more generally useful option might be something like `depth`, where with some external knowledge of the shape of the object hierarchy, a system could limit recursion to some specified depth, `direct` being depth 0 or 1 depending on how the underlying recursive query gem likes to count.  Don't think that is necessary for the immediate use case, though

lib/policy_machine/version.rb
ejinotti-mdsol: so 3.1.0 sounds correct to you? last version bump (not in this file, but in general) was 3.0.1 => 3.0.2 for the addition of the `pluck` method.

DMcKinnon-mdsol: We probably should have bumped `3.1.0` last time. I'm fine doing it this time.

DMcKinnon-mdsol (approved):

PR #184, opened 2021-03-03
[MCC-715209] Multi-Operation Accessible Objects

ejinotti-mdsol:
@mdsol/team04 adds the function `PolicyMachine#accessible_objects_for_operations` which takes multiple operations and returns a map of them to objects accessible through them.  It is only implemented for the `ActiveRecord` adapter and only for directly associated objects.  Did not bother with indirect because:
- It was going to be complicated and require another custom SQL query similar to the one added here for opsets/operations.
- We only need direct for our purposes.


lib/policy_machine_storage_adapters/active_record/postgresql.rb
```diff
@@ -2,6 +2,72 @@
...
+              INNER JOIN accessible_operations
+                ON accessible_operations.child_id = assignments.parent_id
+            )
+          )
+
+          SELECT DISTINCT accessible_operations.operation_set_id, ops.unique_identifier
+          FROM accessible_operations
+          JOIN policy_elements ops
+            ON ops.id = accessible_operations.child_id
+          WHERE ops.unique_identifier IN (?)
```
fpcyan, line 34: Does this do additional filtering not done in L18?
also, hello.

fpcyan: oooo nvm youre traversing the operation_set graph

ejinotti-mdsol: lol howdy, i didn't realize this was a public gem 😉 .. yes, this is just a slight variation of `with_accessible_operation` down below

spec/policy_machine_storage_adapters/active_record_spec.rb
```diff
@@ -311,6 +311,93 @@
...
+              #   create.to_s => [object_6.stored_pe, object_7.stored_pe],
+              #   paint.to_s => [object_6.stored_pe, object_7.stored_pe],
+              # }
+              expect(result.keys).to contain_exactly(create.to_s, paint.to_s)
+              expect(result[create.to_s]).to contain_exactly(object_6.stored_pe, object_7.stored_pe)
+              expect(result[paint.to_s]).to contain_exactly(object_6.stored_pe, object_7.stored_pe)
+            end
+
+            it 'works with filters' do
+              filters = { user_attributes: { color: color_1.color } }
```
kjerinsky-mdsol: Neutral question: Should `filters` and possibly other parameters such as `operations` be initialized by `let` blocks?

kjerinsky-mdsol (approved):
LGTM. I really appreciate the code comments! Makes reviewing much easier to follow, good job 👍

spec/policy_machine_storage_adapters/active_record_spec.rb
ejinotti-mdsol: yea, for `filters` i guess i was just too lazy to make another `context` and putting it at the top of this context above another unrelated example seemed bad.  i'll make a context tho.

with regard to `operations`, do you mean the two item array `[create, paint]`?  in that case i think it's better to just have them typed explicitly in each example.

kjerinsky-mdsol: I was thinking the same thing for the array. One positive is it prevents unintentional values but the negative is needing to scroll a bit to reference the values.

lib/policy_machine/version.rb
```diff
@@ -1,3 +1,3 @@
 class PolicyMachine
-  VERSION = "3.1.0"
+  VERSION = "3.1.1"
```
ejinotti-mdsol: opinions on version number?  couldn't decide between `3.1.1` and `3.2.0`.

CHANGELOG.md
```diff
@@ -1,5 +1,10 @@
 # Changelog
 
+## 3.1.1
+* Added `PolicyMachine#accessible_objects_for_operations` for finding a map of accessible
+  object for multiple operations. Only implemented in `ActiveRecord` and only for
```
cmcinnes-mdsol: nitpick `object` => `objects`

lib/policy_machine_storage_adapters/active_record.rb
```diff
@@ -778,14 +803,38 @@ def accessible_objects(user_or_attribute, operation, options = {})
       end
     end
 
+    # Returns a map of operation names to the user_or_attribute's accessible objects via each operation
+    def accessible_objects_for_operations(user_or_attribute, operations, options = {})
+      unless options[:direct_only]
+        raise ArgumentError, 'Functionality for indirect objects is not yet implemented!'
```
cmcinnes-mdsol, line 809: maybe `raise NotImplementedError` ? though this is triggered by a particular arg so maybe ArgumentError is better

lib/policy_machine_storage_adapters/active_record.rb
```diff
@@ -833,6 +882,46 @@ def objects_for_user_or_attribute_and_operation(user_or_attribute, operation, op
...
+    def objects_for_user_or_attribute_and_operations(user_or_attribute, operations, options)
+      return {} if operations.empty?
+      associations = associations_for_user_or_attribute(user_or_attribute, options)
+
+      # the operation set IDs and the object attribute IDs they connect to
+      # from the user_or_attribute's associations
+      opset_ids_to_objattr_ids = associations.pluck(
+        :operation_set_id,
+        :object_attribute_id
+      ).reduce(Hash.new { |h, k| h[k] = [] }) do |acc, (opset_id, objattr_id)|
```
cmcinnes-mdsol: can save a line with `each_with_object`

lib/policy_machine_storage_adapters/active_record/postgresql.rb
```diff
@@ -1,7 +1,77 @@
...
+        #
+        result = connection.execute(sanitized_query)
+
+        # return a hash like:
+        # {
+        #   'operation1' => [123],
+        #   'operation2' => [123, 789],
+        #   'operation3' => [789],
+        # }
+        result.to_a.reduce(Hash.new { |h, k| h[k] = [] }) do |acc, row|
```
cmcinnes-mdsol: same thing here with `each_with_object`

lib/policy_machine_storage_adapters/active_record.rb
```diff
@@ -33,7 +33,32 @@ def self.const_missing(name)
     end
 
     def self.load_db_adapter!
-      require_relative("active_record/#{PolicyElement.configurations[Rails.env]['adapter']}")
+      require_relative("active_record/#{db_config['adapter']}")
+    end
+
+    def self.db_config
+      @_db_config ||= begin
```
cmcinnes-mdsol, line 40: nonblocking but i've never really liked the `@_<memoized>` pattern with the underscore.

lib/policy_machine/version.rb
cmcinnes-mdsol: i vote `3.2.0`, we bumped minor for the `direct_only` thing

kjerinsky-mdsol: The changelog seems to be inconsistent about when an added method merits a minor bump vs point bump. In a vacuum, I think this would receive a minor version change for updating the outward facing API and reserve points for small tweaks?

That being said, it seems relatively arbitrary.

ejinotti-mdsol: Thanks, yea the "minor version for change in outward facing API" makes sense to me.  Will go with `3.2.0`.

policy_machine.gemspec
```diff
@@ -20,12 +20,11 @@ Gem::Specification.new do |s|
   s.add_dependency('listen')
 
   # Only required in ActiveRecord mode
-  s.add_dependency('activerecord-import', '~> 0.12')
-    # Only required for mysql
-    s.add_dependency('mysql2')
-    # Only required for postgres
-    s.add_dependency('pg', '< 1.0.0')
-    s.add_dependency('activerecord-hierarchical_query', '~> 0.0')
```
cmcinnes-mdsol, line 28: what was this?

cmcinnes-mdsol (commented):
thx @ejinotti-mdsol just some small Qs

lib/policy_machine_storage_adapters/active_record.rb
ejinotti-mdsol: Yea, to me `ArgumentError` feels more appropriate here?  Also, if u look in `lib/policy_machine.rb` in all the cases there where they should be using `NotImplementedError` they instead use `NoMethodError`... which is why I went with `NoMethodError` there.

lib/policy_machine_storage_adapters/active_record.rb
ejinotti-mdsol: yea, i am kinda an `each_with_object` hater.. altho, i guess since it is now part of Ruby itself rather than just a Rails hack perhaps i can put away my hate.

lib/policy_machine_storage_adapters/active_record.rb
ejinotti-mdsol: was mainly just concerned with getting zonked by something else using a generic class instance var name like `@config` or `@db_config`.

policy_machine.gemspec
ejinotti-mdsol: https://github.com/take-five/activerecord-hierarchical_query

it's basically a gem that adds DSL-like functions to AR to generate recursive queries.  it is not actually used as they have instead just written out the queries in SQL directly.  perhaps they were trying to use it as some point, but then gave up and forgot to remove it from dependencies and the one `require`.

lib/policy_machine_storage_adapters/active_record/postgresql.rb
```diff
@@ -1,7 +1,77 @@
-require 'active_record/hierarchical_query' # via gem activerecord-hierarchical_query
-
 module PolicyMachineStorageAdapter
   class ActiveRecord
+    class PolicyElement
+
+      # Given a list of operation set ids and a list of operations
+      # Returns a map of operations to the given operation set ids that contain them
+      def self.filtered_operation_set_ids_by_operation(operation_set_ids, operations)
```
cmcinnes-mdsol, line 7: @ejinotti-mdsol i don't _think_ this will return operations that either don't exist or exist but aren't in an operation set. not sure how we want to handle these but i think the wrapper function should return empty arrays for these operations.
right now 
`PM.objects_for_user_or_attribute_and_operations(operator_pe, ['read', 'frog'], direct_only: true)` 
will return
`{ 'read' => [obj, obj] }` and i think it should return `{ 'read' => [obj, obj], 'frog' => [] }` 

if you agree its prob simpler to have the wrapper function just add the operation keys if they're not there, rather than changing the sql here.

ejinotti-mdsol: ah, good observation.  will add a spec for this.

CHANGELOG.md
```diff
@@ -1,5 +1,10 @@
 # Changelog
 
+## 3.2.0
+* Added `PolicyMachine#accessible_objects_for_operations` for finding a map of accessible
```
mszenher, line 4: Thanks for using PolicyMachine language (object) to describe these concepts rather than Dalton language (operable).

lib/policy_machine.rb
```diff
@@ -292,6 +292,14 @@ def accessible_objects(user_or_attribute, operation, options = {})
     end
   end
 
+  def accessible_objects_for_operations(user_or_attribute, operation, options = {})
```
mszenher: `operations` is plural in the method name but the param `operation` is singular.  Which should it be?

ejinotti-mdsol: oops, this was just a typo, should be plural.  prob because the singular is used often and i hit tab for autocomplete too hastily in my editor.  thanks for catching it!

lib/policy_machine_storage_adapters/active_record.rb
```diff
@@ -1020,9 +1113,32 @@ def ignore_case_applies?(ignore_case, key)
...
+    #   - a PolicyMachineStorageAdapter::ActiveRecord::Operation instance
+    #   - a PM::Operation instance
+    # and returns its identifier (aka name)
+    def operation_identifier(operation)
+      operation.try(:unique_identifier) || operation.to_s
+    end
+
+    # gives the theoretical prohibition identifier (name) for a given operation
+    # NOTE: does not confirm that this prohibition exists like the funcs below
+    def prohibition_id(operation)
```
mszenher: For consistency, how about calling this `prohibition_identifier` ?

ejinotti-mdsol: sure 👍

lib/policy_machine_storage_adapters/active_record.rb
```diff
@@ -778,14 +803,46 @@ def accessible_objects(user_or_attribute, operation, options = {})
...
+    def accessible_objects_for_operations(user_or_attribute, operations, options = {})
+      unless options[:direct_only]
+        raise ArgumentError, 'Functionality for indirect objects is not yet implemented!'
+      end
+
+      # convert to operation names if operation instances given
+      operations = operations.map { |o| operation_identifier(o) }
+
+      # default objects for each operation to empty list
+      permitted_map = operations.map { |o| [o, []] }.to_h
```
mszenher: This may be a premature optimization, but is there any opportunity for parallelization here?  Either parallelizing the calls to `objects_for_user_or_attribute_and_operations` and/or parallelizing the permitted vs. prohibitions computation.

lib/policy_machine_storage_adapters/active_record/postgresql.rb
mszenher: Fiona!!!!!

lib/policy_machine_storage_adapters/active_record.rb
ejinotti-mdsol: Yes, we could do threaded calls to `objects_for_user_or_attribute_and_operations` similar to what we added to Dalton priv calculations.  However this second call to `objects_for_user_or_attribute_and_operations` for the prohibitions is only added for completeness here.  Dalton will only ever call this function with `ignore_prohibitions=true` because it has to do the permitted vs prohibited calculations itself due to it storing the right side (object side) of the graph in Neptune.

I have a draft PR open there where you can see this.  I didn't add threading there either because I think when we did add that, it was more trouble that it was worth: lost trace ids on logs, DB connection problems, etc.  Still open for debate though.

lib/policy_machine_storage_adapters/active_record.rb
```diff
@@ -778,14 +803,46 @@ def accessible_objects(user_or_attribute, operation, options = {})
...
+      permitted_map.merge!(
+        objects_for_user_or_attribute_and_operations(user_or_attribute, operations, options)
+      )
+
+      return permitted_map if options[:ignore_prohibitions]
+
+      prohibitions = prohibitions_for(operations).map { |p| operation_identifier(p) }
+      prohibitted_map = objects_for_user_or_attribute_and_operations(
```
cmcinnes-mdsol: `prohibitted_map` => `prohibited_map`

lib/policy_machine_storage_adapters/active_record.rb
cmcinnes-mdsol: yea i def don't wanna add threading to this gem. that being said i think we can do this in few queires: if `:ignore_prohibitions` is false i think we can run:

```
     prohibited_and_permitted_map = objects_for_user_or_attribute_and_operations(
        user_or_attribute,
        operations + prohibitions,
        options.except(:filters)
      )
```
this would change the filtering logic after that but would cut down the number of db queries.

all that being said i hear ya that dalton is only gna call it with `ignore_prohibitions=true`

lib/policy_machine_storage_adapters/active_record.rb
cmcinnes-mdsol: do you mean `ArgumentError` or `NoMethodError`? either way im fine as is w `ArgumentError`

lib/policy_machine_storage_adapters/active_record.rb
ejinotti-mdsol: ah nice, yea with adding the ops and prohibs together as you suggest we don't even need to consider threading.. perhaps this could be done from dalton also.  guess it is an advantage of building this map.

ejinotti-mdsol: in fact, we could even use this to avoid threading on privilege/show i think

cmcinnes-mdsol: oh hell yea i would looove to rip threading out of priv show

lib/policy_machine_storage_adapters/active_record.rb
ejinotti-mdsol: i mean in `lib/policy_machine.rb` i went with `NoMethodError` for when this function `accessible_objects_for_operations` is not implemented on an adapter just because it is used in the rest of that file, even though i think `NotImplementedError` would be more correct there.

`lib/policy_machine.rb` is basically a higher level wrapper than this active record adapter file we are in here.  it defines the functions of the `PolicyMachine` class and then tries to send them on whatever adapter is configured.  if the particular adapter does not have an implementation of the function, it raises `NoMethodError`.

cmcinnes-mdsol: ah gotcha didn't realize you were referring to that other file, resolving

lib/policy_machine_storage_adapters/active_record.rb
```diff
@@ -778,14 +803,51 @@ def accessible_objects(user_or_attribute, operation, options = {})
...
+      op_and_prohib_names = operation_names + prohibition_names
+      accessible_map.merge!(
+        objects_for_user_or_attribute_and_operations(user_or_attribute, op_and_prohib_names, options)
+      )
+
+      operation_names.each do |operation_name|
+        prohibition_name = prohibition_identifier(operation_name)
+        prohibited_objects = accessible_map[prohibition_name] || []
+        accessible_map[operation_name] -= prohibited_objects
+        accessible_map.delete(prohibition_name)
```
cmcinnes-mdsol, line 837: i might be reading this wrong but i think with this approach if i have `read` on `dog` and `frog` and `~read` on `dog` the accessible map will be: 
```
{ 'read' => ['frog'] }
```
i think there's a problem with this (similar to an observation you made previously for priv#report stuff) because a user can be assigned on an all collection but prohibited on a single operable. so if i had `read` on `all_frogs` and `~read` on `tree_frog` this fn as is would return
```
{ 'read' => ['all_frogs'] }
```
meaning the information on the individual prohibition is lost.

i think we can get around this by just letting the caller of the function (dalton) sort it out, so having this function return:
for the first example 
```
{ 'read' => ['dog', 'frog'], '~read' => ['dog'] }
```
and for the second 
```
{ 'read' => ['all_frogs'], '~read' => ['tree_frog'] }
```

ejinotti-mdsol: This is PM, not Dalton.  PM has no knowledge of the magical meaning of all collections, and Dalton itself will always call this func with `ignore_prohibitions=true`.  This prohibitions calculation is just here for completeness.

In the PR integrating this into Dalton, you can see that the logic for this (including all collections I hope) is handled by the `calculate_accessible_uris` func that you wrote.

cmcinnes-mdsol: ah yes right the plan is to call this with `operations + prohibitions` 👍

cmcinnes-mdsol (approved):

cmcinnes-mdsol:
@mdsol/team04 merging in 30 mins unless objections

PR #186, opened 2021-04-29
[MCC-752251] Fix Policy Machine constant resolution

njoshi-mdsol:
Background-
An error occurs occassionally in all environments where PM receives a constant incorrectly fails and to resolve the same.
Following error is thrown:
Returning an error with status {:status=>:internal_server_error, :message=>“Error message stored in logs”}, NameError: uninitialized constant PolicyMachineStorageAdapter::ActiveRecord::PolicyElement::Assignment

**Constant Received**
PolicyMachineStorageAdapter::ActiveRecord::PolicyElement::Assignment
**Actual class**
PolicyMachineStorageAdapter::ActiveRecord::Assignment

This PR aims to provide the absolute scope for constant resolution.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,8 @@
 # Changelog
 
+## 3.3.1
+* Fix intermittent incorrect constant resolution.
+
 ## 3.3.0
 * Added `PolicyMachine#all_operations_for_user_or_attr_and_objs_or_attrs` for finding a map of all operations
   between given objects and a given user.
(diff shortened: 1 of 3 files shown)


ejinotti-mdsol:
Thanks @njoshi-mdsol , this looks like a good solution.  Did you find a way to reproduce the error so as to know that this works?

njoshi-mdsol:
@ejinotti-mdsol Thanks Ed. Nah, error did not occur with this change. I had this change deployed for a long time and did not see anything failing with the same error. Probably will think more on how to validate this change.

ejinotti-mdsol (approved):

PR #187, opened 2021-05-19
[MCC-773364] Fix filter+prohibition logic in `accessible_objects_for_operations`

ejinotti-mdsol:
@mdsol/team04 this is just for correctness.  In Dalton we only use this with `ignore_prohibitions=true` because reasons.

### DIFF
--- a/lib/policy_machine_storage_adapters/active_record.rb
+++ b/lib/policy_machine_storage_adapters/active_record.rb
@@ -983,7 +983,7 @@ module PolicyMachineStorageAdapter
       #   'write' => [opset1_id],
       #   'delete' => [opset2_id],
       # }
-      operations_to_opset_ids = opset_id_operation_rows.to_a.each_with_object(
+      operations_to_opset_ids = opset_id_operation_rows.each_with_object(
         Hash.new { |h, k| h[k] = [] }
       ) do |row, acc|
         acc[row['unique_identifier']] << row['operation_set_id']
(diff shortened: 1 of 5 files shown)


lib/policy_machine_storage_adapters/active_record.rb
```diff
@@ -870,20 +870,23 @@ def accessible_objects_for_operations(user_or_attribute, operations, options = {
...
       prohibition_names = prohibitions_for(operation_names).map { |p| operation_identifier(p) }
-      op_and_prohib_names = operation_names + prohibition_names
-      accessible_map.merge!(
-        objects_for_user_or_attribute_and_operations(user_or_attribute, op_and_prohib_names, options)
-      )
+
+      accessible_map.merge!(objects_for_user_or_attribute_and_operations(
+        user_or_attribute,
+        prohibition_names,
+        options.except(:filters)
```
kjerinsky-mdsol, line 888: What are filters?

kjerinsky-mdsol (approved):

lib/policy_machine_storage_adapters/active_record.rb
ejinotti-mdsol: they are basically `where`s that get applied to the calculation so that you can calculate, for example, only access that goes through a specific user attribute (role assignment).

apejovich (approved):
LGTM

PR #189, opened 2021-10-11
[MCC-826703] Add option for object fields

kjerinsky-mdsol:
Changes:

`#accessible_objects`

* Added `fields` to `#build_accessible_object_scope`
* ~There are a few `SELECT "policy_elements".*` invoked which are not addressed. Instead, a simple `slice` after the fact.~
  * ~A more traditional `pluck` can be used with some additional work. I'm not certain at this time if there is a performance need for it.~
  * ~I didn't spot any specs for `options[:includes]` which I think should be added if we want to ensure pluck works for all situations. Maybe we should add specs here regardless?~
* Uses pluck where possible to minimize unnecessary instantiated objects.
* Added specs for `includes` option.

## SQL comparison: 

<details>
  <summary>Before</summary>

```SQL
D, [2021-10-12T14:50:17.727679 #14279] DEBUG -- :    (0.8ms)  SELECT "policy_elements"."id" FROM "policy_elements" WHERE (          id IN (
            WITH RECURSIVE assignments_recursive AS (
              (
                SELECT child_id, parent_id
                FROM assignments
                WHERE parent_id in (1)
              )
              UNION ALL
              (
                SELECT assignments.child_id, assignments.parent_id
                FROM assignments
                INNER JOIN assignments_recursive
                ON assignments_recursive.child_id = assignments.parent_id
              )
            )

            SELECT assignments_recursive.child_id
            FROM assignments_recursive
          )
)
D, [2021-10-12T14:50:17.730365 #14279] DEBUG -- :   PolicyMachineStorageAdapter::ActiveRecord::PolicyElementAssociation Load (0.4ms)  SELECT "policy_element_associations".* FROM "policy_element_associations" WHERE "policy_element_associations"."user_attribute_id" IN ($1, $2, $3, $4)  [["user_attribute_id", 2], ["user_attribute_id", 3], ["user_attribute_id", 4], ["user_attribute_id", 1]]
D, [2021-10-12T14:50:17.732689 #14279] DEBUG -- :    (1.2ms)  SELECT "policy_element_associations"."object_attribute_id" FROM "policy_element_associations" WHERE "policy_element_associations"."user_attribute_id" IN ($1, $2, $3, $4) AND (          operation_set_id IN (
            WITH RECURSIVE accessible_operations AS (
              (
                SELECT
                  child_id,
                  parent_id,
                  parent_id AS operation_set_id
                FROM assignments
[...]

PR #190, opened 2021-10-27
[MCC-833873] Add accessible_objects_for_operations function

kjerinsky-mdsol:
This adds a new PostgreSQL function with optimized performance for using `accessible_objects_for_operations` with a single field.

In our test environments with highly assigned users (50k+) we were previously seeing 3-5 second responses. Using `EXPLAIN (ANALYZE)` on a single CTE version we were getting ~ 570 ms. This function has been getting ~ 370-400ms. I'll try to add real benchmarks tomorrow.

Related #189

## Metrics

50K+ assigned users

~ 90% faster

### accessible_objects

Single run

| method version | user     | system    | total     | real      | faster |
| ---            | ---:     | ---:      | ---:      | ---:      | ---:   |
| no pluck       | 5.015483 | 0.2036523 | 5.2191353 | 12.914141 |        |
| pluck          | 3.386614 | 0.082487  | 3.469101  | 10.078440 | 22.0%  |
| pg function    | 0.050423 | 0.013683  | 0.064106  | 0.760845  | 94.1%  |

Multiple runs (5)

| method version | user      | system   | total     | real      | faster |
| ---            | ---:      | ---:     | ---:      | ---:      | ---:   |
| no pluck       | 26.263784 | 0.664979 | 26.928763 | 63.671396 |        |
| pluck          | 16.306010 | 0.331738 | 16.637748 | 50.099469 | 21.3%  |
| pg function    | 0.295134  | 0.051871 | 0.347005  | 3.882921  | 94.0%  |

### accessible_objects_for_operations

| method version | user     | system    | total     | real     | faster |
| ---            | ---:     | ---:      | ---:      | ---:     | ---:   |
| no pluck       | 3.24225  | 0.1424005 | 3.3846505 | 3.944331 |        |
| pluck          | 1.453283 | 0.031504  | 1.484787  | 2.106972 | 46.6%  |
| pg function    | 0.053199 | 0.010115  | 0.063314  | 0.484788 | 87.7%  |

Multiple runs (5)

| method version | user      | system   | total     | real      | faster |
| ---            | ---:      | ---:     | ---:      | ---:      | ---:   |
| no pluck       | 20.315424 | 0.804072 | 21.119496 | 23.865290 |        |
| pluck          | 8.327800  | 0.211137 | 8.538937  | 11.498632 | 51.8%  |
| pg function    | 0.301327  | 0.060720 | 0.362047  | 2.441097  | 89.8%  |


lib/policy_machine_storage_adapters/active_record.rb
```diff
@@ -871,6 +871,11 @@ def accessible_objects_for_operations(user_or_attribute, operations, options = {
         raise ArgumentError, 'Functionality for indirect objects is not yet implemented!'
       end
 
+      # Performance optimized function for PostgreSQL
+      if postgres? && options[:fields]&.one?
```
cmcinnes-mdsol: i don't love the way we're checking for postgres in a general `active_record` adapter class. ideally i think the function would be in `lib/policy_machine_storage_adapters/active_record/postgresql.rb`.

that said i see we're doing this adapter checking elsewhere in this file so i'm ok with leaving it for the future

cmcinnes-mdsol: altho can we make the two places use the same helper function:
https://github.com/mdsol/the_policy_machine/blob/develop/lib/policy_machine_storage_adapters/active_record.rb#L1320 ?
not sure which way of checking for postgres is better but lets just use one

lib/policy_machine/version.rb
```diff
@@ -1,3 +1,3 @@
 class PolicyMachine
-  VERSION = "3.3.4"
+  VERSION = "3.3.5"
```
cmcinnes-mdsol: this feels like it should be a minor version bump

lib/migrations/accessible_objects_for_operations_function.rb
```diff
@@ -0,0 +1,150 @@
...
+    SQL
+  end
+
+  def down
+    execute 'DROP FUNCTION IF EXISTS pm_accessible_objects_for_operations' if postgres?
+  end
+
+  private
+
+  def postgres?
```
cmcinnes-mdsol: also here: do migrations have access to functions defined elsewhere? maybe the `postgres?` fn could live in some utils file that is required here and in the activerecord storage adapter class

lib/policy_machine_storage_adapters/active_record.rb
kjerinsky-mdsol: Postgres would still use `accessible_objects_for_operations` when there isn't one field requested which might make it difficult to share code otherwise. I could make the function more complex to handle more scenarios but we don't personally need that feature.

kjerinsky-mdsol: I'm not sure why the code in your link is using the `Assignment` class for the connection but I can try abstracting it.

cmcinnes-mdsol: > I could make the function more complex to handle more scenarios but we don't personally need that feature.

ya i'm ok with kicking that can for now. i think there's sort of an open of question about how much we'll use this gem. if we start using it more then i think there's lots of cleanup we could do.

lib/policy_machine_storage_adapters/active_record/postgresql.rb
```diff
@@ -74,6 +74,40 @@ def self.operations_for_operation_sets(operation_set_ids, operation_names = nil)
...
+        sanitized_query = sanitize_sql_for_assignment([
+          query,
+          user_id,
+          ActiveRecord.array_to_string(operation_names),
+          field,
+          JSON.dump(filters)
+        ])
+
+        # [
+        #   { 'unique_identifier' => 'op1', 'objects' => '{obj1,obj2,obj3}' },
```
cmcinnes-mdsol, line 92: thanks for these comments they're rly helpful. curious why the postgres fn is returning this string of `{obj1,obj2,obj3}`, is that standard?

lib/policy_machine_storage_adapters/active_record.rb
```diff
@@ -1077,6 +1082,43 @@ def build_accessible_object_scope(associations, options = {})
...
+
+      operation_names = operations.map do |o|
+        name = operation_identifier(o)
+        # Store prohition name if not ignoring prohibitions
+        prohibition_names << prohibition_identifier(name) unless options[:ignore_prohibitions]
+        name
+      end
+
+      all_names = operation_names + prohibition_names
+      accessible_map = all_names.map { |o| [o, []] }.to_h
```
cmcinnes-mdsol:
```suggestion
      accessible_map = all_names.index_with([])
```
part of activesupport but this gem has activesupport as a dependency so we should be covered

cmcinnes-mdsol: also, does this block mean that the function won't return the operation name key if there are no accessible object uris?

lib/policy_machine_storage_adapters/active_record.rb
```diff
@@ -1077,6 +1082,43 @@ def build_accessible_object_scope(associations, options = {})
...
+      # improvement over many ad hoc ActiveRecord calls.
+      if options[:filters] && options[:ignore_prohibitions] != true
+        result = PolicyElement.accessible_objects_for_operations(user_id, operation_names, options)
+        accessible_map.merge!(result)
+        result = PolicyElement.accessible_objects_for_operations(user_id, prohibition_names, options.except(:filters))
+        accessible_map.merge!(result)
+      else
+        result = PolicyElement.accessible_objects_for_operations(user_id, all_names, options)
+        accessible_map.merge!(result)
+      end
```
cmcinnes-mdsol: maybe something like:

```suggestion
      postgres_result = 
        if options[:filters] && options[:ignore_prohibitions] != true
          perms_map = PolicyElement.accessible_objects_for_operations(user_id, operation_names, options)
          prohibs_map = PolicyElement.accessible_objects_for_operations(user_id, prohibition_names, options.except(:filters))
          perms_map.merge(prohibs_map)
        else
          PolicyElement.accessible_objects_for_operations(user_id, all_names, options)
        end
      
      accessible_map.merge!(postgres_result)
```
?

lib/policy_machine_storage_adapters/active_record.rb
```diff
@@ -1077,6 +1082,43 @@ def build_accessible_object_scope(associations, options = {})
...
+
+      return accessible_map if options[:ignore_prohibitions]
+
+      operation_names.each do |operation_name|
+        prohibition_name = prohibition_identifier(operation_name)
+        prohibited_objects = accessible_map[prohibition_name] || []
+        accessible_map[operation_name] -= prohibited_objects
+        accessible_map.delete(prohibition_name)
+      end
```
cmcinnes-mdsol: i know this is never used in dalton so not a big deal but can save a line with:

```suggestion
      operation_names.each do |operation_name|
        prohibition_name = prohibition_identifier(operation_name)
        prohibited_objects = accessible_map.delete(prohibition_name) || []
        accessible_map[operation_name] -= prohibited_objects
      end
```

lib/policy_machine_storage_adapters/active_record.rb
cmcinnes-mdsol: if so, would it be easy to change the function to just return an empty string when there are no accessible object uris? not a big deal but that could simplify the code here a bit.

lib/policy_machine_storage_adapters/active_record/postgresql.rb
```diff
@@ -74,6 +74,40 @@ def self.operations_for_operation_sets(operation_set_ids, operation_names = nil)
...
+        #   { 'unique_identifier' => 'op2', 'objects' => '{obj1,obj2,obj3}' },
+        # ]
+        result = connection.execute(sanitized_query).to_a
+
+        # {
+        #    'op1' => [{ :field => 'obj1' }, { :field => 'obj2' }, { :field => 'obj3' }],
+        #    'op2' => [{ :field => 'obj2' }, { :field => 'obj3' }, { :field => 'obj4' }],
```
cmcinnes-mdsol: this is kinda a funky thing to return, is the invoking function expect this? does it need to?

lib/policy_machine_storage_adapters/active_record/postgresql.rb
```diff
@@ -74,6 +74,40 @@ def self.operations_for_operation_sets(operation_set_ids, operation_names = nil)
...
+        # }
+        result.map do |h|
+          objects = ActiveRecord.string_to_array(h['objects'])
+          value = objects.map { |o| { field => o }}
+          [h['unique_identifier'], value]
+        end.to_h
```
cmcinnes-mdsol: i think `each_with_object` would be slightly more efficient here as it would only do one loop
```suggestion
        result.each_with_object({}) do |result_hash, output|
          objects = ActiveRecord.string_to_array(result_hash['objects'])
          key = result_hash['unique_identifier']
          output[key] = objects.map { |o| { field => o }}
        end
```

cmcinnes-mdsol (commented):
looking good! a few ruby object manipulation suggestions

lib/policy_machine/version.rb
kjerinsky-mdsol: Unresolving because @ejinotti-mdsol might disagree.

lib/policy_machine_storage_adapters/active_record/postgresql.rb
kjerinsky-mdsol: Yes, this is the format that postgres represents arrays and JSON https://www.postgresql.org/docs/current/arrays.html#ARRAYS-INPUT.

Incoming commit will use `PG::TextDecoder::Array` which uses C extensions. Should be better than my DIY version.

lib/policy_machine/version.rb
ejinotti-mdsol: yes, i do.  this change does not affect the gem's API at all so seems like tiny version to me.  in other words, should be invisible to anyone using the gem.

lib/policy_machine_storage_adapters/active_record.rb
kjerinsky-mdsol: That shape doesn't seem to be inline with the existing specs to me.

lib/policy_machine_storage_adapters/active_record/postgresql.rb
kjerinsky-mdsol: No and I'm not a fan of it either. We can update our consuming end if we want to take it out.

lib/policy_machine_storage_adapters/active_record.rb
kjerinsky-mdsol: 24b4a9d

lib/policy_machine_storage_adapters/active_record/postgresql.rb
kjerinsky-mdsol: 24b4a9d

lib/policy_machine_storage_adapters/active_record.rb
kjerinsky-mdsol: 24b4a9d

lib/policy_machine/version.rb
kjerinsky-mdsol: I agree, though that won't be the case if we change the shape of the return https://github.com/mdsol/the_policy_machine/pull/190#discussion_r742319869

lib/policy_machine_storage_adapters/active_record/postgresql.rb
```diff
@@ -417,5 +447,12 @@ def self.apply_include_condition(scope: , key: , value: , klass: )
         end
       end
     end
+
+    private
+
+    # Convert array to '{element1,element2,element3}'
+    def self.array_to_string(array)
```
cmcinnes-mdsol: this is no longer need right?

lib/policy_machine/version.rb
cmcinnes-mdsol: i also don't think it's really invisible even if we don't change the shape of the return because it requires generating a new migration?

lib/policy_machine_storage_adapters/active_record/postgresql.rb
ejinotti-mdsol: i think maybe this comment is deceiving you?

this function originally returned a hash where the keys were operation names and the values were the accessible objects for each operation.

after the pluck changes, i believe it returns similar except in place of the objects are now hashes with keys/values of the requested fields (which i believe is what this comment is supposed to represent?)?

lib/policy_machine_storage_adapters/active_record/postgresql.rb
kjerinsky-mdsol: yep, forgot to remove

lib/policy_machine_storage_adapters/active_record/postgresql.rb
kjerinsky-mdsol: 24b4a9d

lib/policy_machine_storage_adapters/active_record/postgresql.rb
kjerinsky-mdsol: ca699ce

lib/policy_machine_storage_adapters/active_record.rb
```diff
@@ -1077,6 +1085,51 @@ def build_accessible_object_scope(associations, options = {})
...
+
+    # Performance optimized function for PostgreSQL
+    def accessible_objects_for_operations_function(user_id, operations, options)
+      prohibition_names = []
+
+      operation_names = operations.map do |o|
+        name = operation_identifier(o)
+        prohibition_names << prohibition_identifier(name) unless options[:ignore_prohibitions]
+        name
+      end
```
ejinotti-mdsol: the old code was using this `prohibitions_for` func to get the prohib names which did a `.where` to grab only those that exist in the DB.. since we are not even using this functionality (we always use `ignore_prohibitions=true`) perhaps we shouldn't change that behavior?  it's also kinda icky to build some second array in the `.map`ping of some first array.

lib/policy_machine_storage_adapters/active_record/postgresql.rb
```diff
@@ -74,6 +74,36 @@ def self.operations_for_operation_sets(operation_set_ids, operation_names = nil)
         # ]
         connection.execute(sanitized_query)
       end
+
+      def self.accessible_objects_for_operations(user_id, operation_names, options)
```
ejinotti-mdsol, line 79: can we toss in a comment like "this func assumes that the fields option will be passed with a single item"?

lib/policy_machine_storage_adapters/active_record.rb
cmcinnes-mdsol: i would also be ok with raising a `NotImplementedError` if `ignore_prohibitions!=true`

lib/policy_machine_storage_adapters/active_record/postgresql.rb
kjerinsky-mdsol: d4b0485

lib/policy_machine_storage_adapters/active_record.rb
kjerinsky-mdsol: d4b0485

lib/policy_machine_storage_adapters/active_record.rb
```diff
@@ -1077,6 +1085,25 @@ def build_accessible_object_scope(associations, options = {})
...
+    def use_accessible_objects_function?(options)
+      PolicyMachineStorageAdapter.postgres? &&
+        options[:direct_only] &&
+        options[:ignore_prohibitions] &&
+        options[:fields]&.one?
+    end
+
+    # Performance optimized function for PostgreSQL
+    def accessible_objects_function(user_id, operation, options)
+      accessible_objects_for_operations_function(user_id, Array.wrap(operation), options).values.flatten
```
cmcinnes-mdsol, line 1097: nice, had never seen this `wrap` function

lib/policy_machine_storage_adapters/active_record.rb
```diff
@@ -1077,6 +1085,25 @@ def build_accessible_object_scope(associations, options = {})
...
+    end
+
+    # Performance optimized function for PostgreSQL
+    def accessible_objects_for_operations_function(user_id, operations, options)
+      operation_names = operations.map { |o| operation_identifier(o) }
+      accessible_map = operation_names.index_with([])
```
cmcinnes-mdsol: this will have all the operation names point to the same array (i just did the same thing in another PR). block implementation will be invoked once for each operation name, so will be a different array for each,.
```suggestion
      accessible_map = operation_names.index_with { [] }
```

spec/policy_machine_storage_adapters/active_record_spec.rb
```diff
@@ -246,6 +246,12 @@
       end
 
       describe 'accessible_objects' do
+        before do
+          allow(PolicyMachineStorageAdapter).to receive(:postgres?).and_return(false)
+          expect_any_instance_of(PolicyMachineStorageAdapter::ActiveRecord)
```
cmcinnes-mdsol: what's this for? `expect`s shouldn't really be in a `before` block. it looks like this one is a real expectation so should go in an example (`it`) block.

spec/policy_machine_storage_adapters/active_record_spec.rb
```diff
@@ -378,6 +384,46 @@
         end
       end
 
+      describe 'accessible_objects_function' do
+        context 'direct only' do
+          before do
+            priv_pm.add_association(color_1, creator, object_7)
+            expect_any_instance_of(PolicyMachineStorageAdapter::ActiveRecord)
```
cmcinnes-mdsol: similar comment to above abt expects not going in befores, but i think this one is just setup so can be `allow_any_instance_of`

spec/policy_machine_storage_adapters/active_record_spec.rb
cmcinnes-mdsol: (same goes for similar `expect_any_instance_of...not_to receive` instance below)

spec/policy_machine_storage_adapters/active_record_spec.rb
cmcinnes-mdsol: (same goes for these `expect_any_instance_of..and_call_original`) in before blocks below

spec/policy_machine_storage_adapters/active_record_spec.rb
```diff
@@ -656,6 +708,105 @@
...
+        end
+
+        context 'when there are directly accessible objects' do
+          before do
+            priv_pm.add_association(color_1, creator, object_6)
+            priv_pm.add_association(color_2, creator, object_7)
+          end
+
+          it 'returns objects accessible via each of multiple given operations' do
+            result = priv_pm.accessible_objects_for_operations(
```
cmcinnes-mdsol, line 765: i would normally do this `result` local as a `let` and put these examples in `context` blocks and then set individual params in the `let`. 
that said i feel fairly ambivalent here and dont think u need to spend the extra time 😄

lib/policy_machine_storage_adapters/active_record.rb
kjerinsky-mdsol: Good catch!

cmcinnes-mdsol (approved):
LGTM, thanks @kjerinsky-mdsol !

lib/policy_machine_storage_adapters/active_record.rb
```diff
@@ -1077,6 +1085,25 @@ def build_accessible_object_scope(associations, options = {})
       end
     end
 
+    def use_accessible_objects_function?(options)
+      PolicyMachineStorageAdapter.postgres? &&
+        options[:direct_only] &&
+        options[:ignore_prohibitions] &&
```
cmcinnes-mdsol: actually @kjerinsky-mdsol i know i just approved but what do you think
```
        options[:direct_only] == true &&
        options[:ignore_prohibitions] == true &&
```

kjerinsky-mdsol: d5f0d2a

ejinotti-mdsol (approved):

PR #192, opened 2021-11-15
[MCC-833873] Fix single pg call for accessible_objects from replica

kjerinsky-mdsol:
PostgreSQL has restrictions on creating temporary tables on replica databases. Adding in the CTE that was used in previous benchmark investigations for use when the database connection a replica.

Benchmarks for 50k user

Single run

| method version | user     | system   | total    | real     | faster |
| ---            | ---:     | ---:     | ---:     | ---:     | ---:   |
| no pluck       | 3.163534 | 0.151665 | 3.315199 | 3.845629 |        |
| pluck          | 1.610845 | 0.035368 | 1.646213 | 2.271648 | 40.1%  |
| pg function    | 0.020938 | 0.012064 | 0.033002 | 0.452649 | 88.2%  |
| cte            | 0.041312 | 0.007820 | 0.049132 | 0.624768 | 83.8%  |

Multiple runs (5)

| method version | user      | system   | total     | real      | faster |
| ---            | ---:      | ---:     | ---:      | ---:      | ---:   |
| no pluck       | 17.658298 | 0.479222 | 18.137520 | 21.150227 |        |
| pluck          | 7.143887  | 0.198251 | 7.342138  | 10.431428 | 50.7%  |
| pg function    | 0.182273  | 0.055765 | 0.238038  | 2.381902  | 88.7%  |
| cte            | 0.163428  | 0.039802 | 0.203230  | 3.082762  | 85.4%  |

### DIFF
--- a/lib/policy_machine_storage_adapters/active_record/postgresql.rb
+++ b/lib/policy_machine_storage_adapters/active_record/postgresql.rb
@@ -80,13 +80,22 @@ module PolicyMachineStorageAdapter
         field = options[:fields].first
         filters = options.dig(:filters, :user_attributes) || {}
 
-        query = sanitize_sql_for_assignment([
-          'SELECT * FROM pm_accessible_objects_for_operations(?,?,?,?)',
-          user_id,
-          PG::TextEncoder::Array.new.encode(operation_names),
-          field,
-          JSON.dump(filters)
-        ])
+        query =
+          if readonly?
+            sanitize_sql_for_assignment([
+              accessible_objects_for_operations_cte(field, filters),
+              user_id,
+              operation_names
+            ])
+          else
+            sanitize_sql_for_assignment([
+              'SELECT * FROM pm_accessible_objects_for_operations(?,?,?,?)',
+              user_id,
+              PG::TextEncoder::Array.new.encode(operation_names),
+              field,
+              JSON.dump(filters)
+            ])
+          end
 
         # [
         #   { 'unique_identifier' => 'op1', 'objects' => '{obj1,obj2,obj3}' },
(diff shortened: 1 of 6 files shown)


lib/policy_machine_storage_adapters/active_record/postgresql.rb
```diff
@@ -80,13 +80,22 @@ def self.accessible_objects_for_operations(user_id, operation_names, options)
...
 
-        query = sanitize_sql_for_assignment([
-          'SELECT * FROM pm_accessible_objects_for_operations(?,?,?,?)',
-          user_id,
-          PG::TextEncoder::Array.new.encode(operation_names),
-          field,
-          JSON.dump(filters)
-        ])
+        query =
+          if readonly?
```
cmcinnes-mdsol, line 84: where is this func coming from?

cmcinnes-mdsol: ignore this comment lol

cmcinnes-mdsol:
@kjerinsky-mdsol we also need to fix the loading issue that leads to rake tasks failing and containers failing when they first spin up

policy_machine.gemspec
```diff
@@ -33,5 +33,5 @@ Gem::Specification.new do |s|
   s.add_development_dependency('byebug')
   s.add_development_dependency('neography', '~> 1.1')
   s.add_development_dependency('database_cleaner')
-  s.add_development_dependency('rails', '~> 6.0')
+  s.add_development_dependency('rails', '~> 6.1')
```
cmcinnes-mdsol, line 36: why the bump? asking cause im wondering if this is indeed a development dependency or actually a runtime dependency

kjerinsky-mdsol: Because `ActiveRecord::Base.connection_db_config.config` is logging from Rails 6.1
> DEPRECATION WARNING: DatabaseConfig#config will be removed in 6.2.0 in favor of DatabaseConfigurations#configuration_hash which returns a hash with symbol keys (called from irb_binding at (irb):20)

Rails 6.0 doesn't have `DatabaseConfigurations#configuration_hash` so either get ahead of it or let it potentially sneak up on us.

kjerinsky-mdsol: That or check for either method.

ejinotti-mdsol: See my other comment

lib/policy_machine_storage_adapters/active_record/postgresql.rb
```diff
@@ -105,6 +114,112 @@ def self.accessible_objects_for_operations(user_id, operation_names, options)
...
+
+        filters.each do |key, value|
+          condition << sanitize_sql_for_assignment(["#{connection.quote_column_name(key)} = ? AND ", value])
+        end
+
+        condition.chomp('AND ') << ')'
+      end
+
+      def self.replica?
+        ::ActiveRecord::Base.connection_db_config.configuration_hash[:replica] == true
```
ejinotti-mdsol: There is already a function for doing some ridiculous rigmarole to handle the ever-changing ActiveRecord API surface for getting a DB config:

https://github.com/mdsol/the_policy_machine/blob/6ff77d173dec2ce1f05ff9a60d8d381f598bd0d7/lib/policy_machine_storage_adapters/active_record.rb#L39-L69

Should probably use it somehow.

policy_machine.gemspec
ejinotti-mdsol: Also, this is mainly for the test Rails app that is created for running specs I think.  If we really cared about this gem we would make it use appraisals for various Ruby/Rails combinations... but, to me, just using the same version as Dalto is fine.

cmcinnes-mdsol (approved):

PR #195, opened 2023-06-14
Ruby 3 support

cmcinnes-mdsol:
almost didnt bother fixing the specs but meh


ejinotti-mdsol (changes requested):
need to bump the version in `lib/policy_machine/version.rb`

kjerinsky-mdsol (approved):

ejinotti-mdsol (approved):

PR #196, opened 2023-08-14
Handle empty extra attributes so that we can "includes" plain old AR assocations to PE objects

cmcinnes-mdsol:
here be dragons

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,8 @@
 # Changelog
 
+## 4.0.6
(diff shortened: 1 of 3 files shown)


ejinotti-mdsol:
what does this let u actually do?

cmcinnes-mdsol:
will show u on the dalton pr just pushing up changes

ejinotti-mdsol (approved):

wdutymdsol (approved):

PR #201, opened 2024-04-24
[MCC-1204770] Improve accessible objects pgsql function performance when filtering

kjerinsky-mdsol:
[MCC-1204770](https://jira.mdsol.com/browse/MCC-1204770)

[Notes](https://github.com/mdsol/kjerinsky-mdsol-notes/tree/notes/2024/mcc/MCC-1189079-investigate-slow-priv)

## Changes:

First statement selects by `DISTINCT`. The `child_id` was duplicated at times in significant numbers. A very marginal difference in ms at most if at all.

Altered the filter statement to be more efficient. In the reproducible example, we reduced an 8+ minute call down to 180 ms.

@mdsol/platform-scalability

### DIFF
--- a/lib/generators/the_policy_machine/update_accessible_objects_filter_performance_generator.rb
+++ b/lib/generators/the_policy_machine/update_accessible_objects_filter_performance_generator.rb
@@ -1,0 +1,12 @@
+module ThePolicyMachine
+  module Generators
+    class UpdateAccessibleObjectsFilterPerformanceGenerator < Rails::Generators::Base
+      source_root File.expand_path('../../../migrations', __FILE__)
+
+      def generate_update_accessible_objects_filter_performance_migration
+        timestamp = Time.now.utc.strftime("%Y%m%d%H%M%S")
+        copy_file('update_accessible_objects_filter_performance.rb', "db/migrate/#{timestamp}_update_accessible_objects_filter_performance.rb")
+      end
+    end
+  end
+end
--- a/lib/migrations/update_accessible_objects_filter_performance.rb
+++ b/lib/migrations/update_accessible_objects_filter_performance.rb
@@ -1,0 +1,297 @@
+class UpdateAccessibleObjectsFilterPerformance < ActiveRecord::Migration[5.2]
+  def up
+    return unless PolicyMachineStorageAdapter.postgres?
+
+    execute <<~SQL.squish
+      CREATE OR REPLACE FUNCTION pm_accessible_objects_for_operations(
+        user_id INT,
+        operations _TEXT,
+        field TEXT,
+        filters JSON DEFAULT '{}'
+      )
+      RETURNS TABLE (
+        unique_identifier varchar(255),
+        objects _varchar
+      ) AS $$
+      DECLARE
+        filter_key TEXT;
+        filter_value TEXT;
+        filter_conditions TEXT = '';
+        user_attribute_table TEXT = 't_user_attribute_ids';
+      BEGIN
+        CREATE TEMP TABLE IF NOT EXISTS t_user_attribute_ids (
+          child_id INT
+        );
+
+        CREATE TEMP TABLE IF NOT EXISTS t_user_attribute_filtered_ids (
+          child_id INT
+        );
+
+        CREATE TEMP TABLE IF NOT EXISTS t_operation_set_ids (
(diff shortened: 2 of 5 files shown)


lib/generators/the_policy_machine/update_accessible_objects_filter_performance_generator.rb
```diff
@@ -0,0 +1,12 @@
+module ThePolicyMachine
+  module Generators
+    class UpdateAccessibleObjectsFilterPerformanceGenerator < Rails::Generators::Base
+      source_root File.expand_path('../../../migrations', __FILE__)
+
+      def generate_update_accessible_objects_filter_performance_migration
```
ccowdrey-mdsol, line 6: Just out of curiosity, why do we do all this migration generation manually instead of the standard `rails gem_name:install:migrations`?

lib/migrations/update_accessible_objects_filter_performance.rb
```diff
@@ -0,0 +1,297 @@
...
+          filter_conditions := replace(filter_conditions, '"', '''');
+
+          EXECUTE format(
+            'INSERT INTO t_user_attribute_filtered_ids ' ||
+            'SELECT ua.child_id ' ||
+            'FROM t_user_attribute_ids ua ' ||
+            'JOIN policy_elements pe ' ||
+            '  ON pe.id = ua.child_id ' ||
+            'WHERE %s',
+            filter_conditions
```
ccowdrey-mdsol, line 87: This seems a lot better than the deleting and NOT IN clause. 👍

lib/migrations/update_accessible_objects_filter_performance.rb
```diff
@@ -0,0 +1,297 @@
...
+        TRUNCATE t_user_attribute_ids;
+        TRUNCATE t_user_attribute_filtered_ids;
+        TRUNCATE t_operation_set_ids;
+        TRUNCATE t_accessible_operations;
+        TRUNCATE t_operation_sets;
+
+        RETURN;
+      EXCEPTION
+        WHEN read_only_sql_transaction THEN
+          RETURN QUERY SELECT * FROM pm_accessible_objects_for_operations_cte(user_id, operations, field, filters);
```
ccowdrey-mdsol, line 153: What is this doing?

lib/generators/the_policy_machine/update_accessible_objects_filter_performance_generator.rb
kjerinsky-mdsol: Simple as this is the pattern the gem has been using since many years ago.

lib/migrations/update_accessible_objects_filter_performance.rb
kjerinsky-mdsol: This exception is raised when attempting to create a temporary table from a physical read replica. It's a weak fallback if the app is placed in read only mode.

ccowdrey-mdsol: Oh, interesting. How often is `pm_accessible_objects_for_operations_cte` used? Does it exist solely for RR fallback?

ccowdrey-mdsol (approved):

cmcinnes-mdsol:
notes for other reviewers: https://github.com/mdsol/kjerinsky-mdsol-notes/tree/notes/2024/mcc/MCC-1189079-investigate-slow-priv (link in desc is broken)

cmcinnes-mdsol (commented):
i don't really mind (because we can't get rid of the gem anyways) but thoughts on just defining this fn in dalton itself and not messing w this gem anymore?

kjerinsky-mdsol:
That would be fine. We would just need to move the specs over or create new ones to test it. Bonus would be that I could add scientist experiments.

cmcinnes-mdsol:
besides that the query LGTM, my main question was answered in your notes: `Attempts to filter inside the first statement did not work well. The CTE was not ordering as desired, even with materialization.` 

i lean toward just defining the func in dalton itself and not messing around w this gem for this but i dont feel too strongly and dont wanna create extra work for you

ejinotti-mdsol:
per discussion earlier today, seems like we go with keeping it in the gem for now.  merging.

ejinotti-mdsol (approved):

PR #202, opened 2024-07-29
[MCC-1241918] Add filter ability to CTE version of left side calculation for readonly mode

kjerinsky-mdsol:
[MCC-1241918](https://jira.mdsol.com/browse/MCC-1241918)

* Adds filter ability to the CTE version of the accessible objects function
* Adds parameter option to use CTE instead so we can run scientist experiments

@mdsol/platform-scalability

### DIFF
--- a/lib/migrations/accessible_objects_for_operations_function_cte.rb
+++ b/lib/migrations/accessible_objects_for_operations_function_cte.rb
@@ -1,0 +1,143 @@
+class AccessibleObjectsForOperationsFunctionCte < ActiveRecord::Migration[5.2]
+  def up
+    return unless PolicyMachineStorageAdapter.postgres?
+
+    execute <<~SQL.squish
+      CREATE OR REPLACE FUNCTION pm_accessible_objects_for_operations_cte(
+        user_id INT,
+        operations _TEXT,
+        field TEXT,
+        filters JSON DEFAULT '{}'
+      )
+      RETURNS TABLE (
+        unique_identifier varchar(255),
+        objects _varchar
+      ) AS $$
+      DECLARE
+        mergejoin TEXT;
+        filter_key TEXT;
+        filter_value TEXT;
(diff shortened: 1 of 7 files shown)


cmcinnes-mdsol:
do we wanna just make these changes in dalton itself and not both with the gem?

kjerinsky-mdsol:
I thought about doing that. Aside from specs (which there aren't many and can be ported) I kept the status quo for now as a hedge. It's nice for the temp table function to have a fallback CTE function to call but I kind of want to see how the experiments go before making the next move. I'm open though.

cmcinnes-mdsol:
fine w me. but yea next move we make lets just move the func into dalton proper, i dont think going back and forth with this gem produces much value and the pr dance is kind of annoying

lib/migrations/accessible_objects_for_operations_function_cte.rb
```diff
@@ -0,0 +1,143 @@
...
+          FOR filter_key, filter_value IN
+            SELECT * FROM json_each(filters)
+          LOOP
+            filter_conditions := filter_conditions || filter_key || ' = ' || filter_value || ' AND ';
+          END LOOP;
+
+          /* Chomp trailing AND */
+          filter_conditions := left(filter_conditions, -4);
+          /* Replace double quotes */
+          filter_conditions := replace(filter_conditions, '"', '''');
```
cmcinnes-mdsol, line 54: this `''''` works? i'd think you'd need escaping

kjerinsky-mdsol: Doubling single quotes is common in most RDBMS I've used. There are other options such as adding `E'\'foo\''`, character functions, formatting functions, etc

cmcinnes-mdsol: roger just checking

